PR #86, opened 2026-05-03
Add support for background tag for window images

raidekoptix:
The goal here is to parse an `img` attribute of the `<nav/>` tag that contains a full path to an image file that will be set on the background of the main story window for added ambiance. The `img` path is determined via an external Lich script for greater user flexibility.


raidekoptix:
After discussion with @sproctor we decided a dedicated `<background/>` window would be better than trying to key off of a `<nav/>` attribute...

Swapped to `<background/>` which accepts two attributes `window=` for targetting specific windows, and `img=` for the file path.

### Examples
| Tag  | Effect |
| ------------- | ------------- |
| `<background window='room' img="something.png"/>` | Sets the `room` window to something.png |
| `<background img="something.png"/>` | Sets the `main` story window (default) to something.png |
| `<background window='main'/>` | Clears the image from the main story window |
| `<background/>` | Clears the main window background |

raidekoptix:
The excitement is getting away from me, but I've added several additional attributes. A full list below:

| Attribute  | Effect |
| ------------- | ------------- |
| `window` | target window name. Missing or blank defaults to main. |
| `img` | background image path. Missing, blank, or 0 clears the background. |
| `mode` | image mode. Supported values: `fill`, `hfill` (default), `wfill`, `full`, `gradient` |
| `opacity` | max image opacity, 0..100. Defaults to 100.
| `start` | gradient start position as percent from the image’s left edge, 0-100% Defaults to 0 |
| `end` | gradient end position as percent from the image’s left edge, 0-100% Defaults to 100 |
| `align` | horizontal image alignment. Supported values: `left`,`center` (default), `right` |
| `valign` | vertical image alignment. Supported values: `top`,`middle` (default), `bottom` |

sproctor (commented):
The data part of it looks good. I would say a background tag without a window name should default to "main" and if the server/script wants to clear all backgrounds, it can do the work of sending a background tag for each window.

I think the nav change does nothing and should be removed. I'll have more comments later, but I ran out of time.

raidekoptix:
Sorry about that! Nav changes weren't supposed to still be in there; I've cleaned them up.

Adjusted the default behavior of `<background/>` to default to clearing the main window.

PR #105, opened 2026-05-23
remember main window position

OSXLich-Doug:
## Summary
- save the main Warlock window X/Y/width/height per connected character
- restore saved bounds after the connected character is known
- keep the existing global width/height startup fallback

## Tested
- ./gradlew :desktopApp:compileKotlin -PiosSkip=true -PlintSkip=true
- ./gradlew :core:ktlintCommonMainSourceSetCheck :desktopApp:ktlintMainSourceSetCheck -PiosSkip=true -PlintSkip=true
- manually verified two characters remember different main window positions

Note: full ./gradlew check was not run locally because this environment does not have Android SDK configuration.


sproctor:
I've changed my dev process a bit. I should document this somewhere. I'm running everything against the master branch now. Can you create a new PR to that branch?

PR #109, opened 2026-05-25
remember main window position

OSXLich-Doug:
## Summary
- save the main Warlock window X/Y/width/height per connected character
- restore saved bounds after the connected character is known
- keep the existing global width/height startup fallback

## Tested
- ./gradlew :desktopApp:compileKotlin -PiosSkip=true -PlintSkip=true
- ./gradlew :core:ktlintCommonMainSourceSetCheck :desktopApp:ktlintMainSourceSetCheck -PiosSkip=true -PlintSkip=true
- manually verified two characters remember different main window positions

Note: no android environment available


OSXLich-Doug:
Quick note -  did not execute full `./gradlew` tests.  I don't keep an android environment installed here.  So only focused testing occurred.

sproctor:
Thanks!

PR #180, opened 2026-06-21
Avoid UI stall during rapid game-text floods (e.g. ;go2)

jasonjaclyn2017:
> **Note from human:** I noticed some slowness that made it nearly unplayable, so Claude and I made some fixes.

## Problem

Running a Lich `;go2` flood (rapid room descriptions) on a **single** character froze the UI and then dumped the whole travel log at once, and typing a command could lag for seconds. The cost was per-line work in the stream buffer that scales with the scrollback size, so it showed up even with one window open.

## Changes (two commits)

**`ComposeTextStream`: coalesce publishes + O(1) eviction**
- Every appended line republished a full O(n) `toList()` copy of the scrollback to the `lines` `StateFlow`, and once the buffer hit `maxLines` it evicted the head with `ArrayList.removeAt(0)` — another O(n) shift. Under a flood, the single work-queue coroutine couldn't keep up, so the conflated `StateFlow` dropped every intermediate value and the UI only repainted after the burst finished.
- `finishedLines`/`cacheLines` are now `ArrayDeque` (O(1) head eviction), and snapshots are coalesced: line ops `trySend` to a `CONFLATED` channel and a single collector publishes one snapshot per *burst* instead of per *line*.

**`StyledString`: O(k) line assembly**
- `plus` folded fragments with `(a + b).toPersistentList()`, copying the whole accumulator each time — O(k²) for a fragment-heavy room line. `substrings` is narrowed to `PersistentList` and appended with `addAll` (structural sharing), making line assembly O(k).

## Behavior

Rendered output, line ordering, name-filter, and component-update logic are unchanged — only the per-line cost (now O(1) amortized) and the publish cadence. The `StateFlow` already conflated intermediate emissions, so nothing visible is lost.

## Testing

- `:core:jvmTest` and `:wrayth:jvmTest` pass; `:compose` compiles clean (no new warnings).
- Verified by hand: flooded a character with a long `;go2` — text now streams in incrementally and typing stays responsive during travel, instead of freezing and dumping the log at the end.

### DIFF
--- a/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/window/ComposeTextStream.kt
+++ b/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/window/ComposeTextStream.kt
@@ -4,6 +4,8 @@ import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.buildAnnotatedString
 import co.touchlab.kermit.Logger
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.launchIn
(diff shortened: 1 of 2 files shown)


sproctor:
Superseded. The `ComposeTextStream` changes here (ArrayDeque + coalesced publish) were reworked and merged in #183, so this went conflicting. The remaining `StyledString` O(k) line-assembly fix has been cherry-picked (with original authorship preserved) onto current master in #184. Closing in favor of #183 + #184 — thanks @jasonjaclyn2017!

sproctor:
Well, the Claude generated message said basically what I would've. I just happened to be working on a very similar set of changes at exactly the same time. The `StyledString` fix is a good catch though! It doesn't have much impact on normal streams, but it should help on long fragmented lines. Thank you!

PR #183, opened 2026-06-21
Text-stream buffer perf, suppress-prompts setting, and shared window-view scaffold

sproctor:
> **Draft / bundled.** This carries three loosely-related changes that are entangled across the same files (`ComposeTextStream.kt` especially), so they couldn't be cleanly separated into individual PRs. It also overlaps open #180 (`perf/stream-render-efficiency`) in the stream-rendering area — opening as a draft so it can be reconciled with #180 and the WIP pieces finished before it's considered for merge.

## What's here

**1. ComposeTextStream buffer performance**
Updates the displayed-line list incrementally instead of rebuilding and re-filtering the whole buffer on every line:
- `ArrayDeque` line buffers — trimming the oldest line at capacity is O(1), not an O(n) array shift.
- Displayed view = an append-only persistent list + a logical start index, so evicting the oldest displayed line (every newline once the buffer fills) is an O(1) start bump with amortized compaction; appends are O(log n).
- Identity-equals `OffsetList` snapshot so the `StateFlow`/Compose-state dedup comparison is O(1) instead of an O(n) content scan.
- Result: per-newline work at capacity goes from ~2-3× O(n) to O(log n).

**2. Suppress-prompts setting**
A client setting that filters prompt lines out of the displayed stream, with desktop and mobile settings toggles (`ConfigSchema`, `ClientSettingRepository`, `WindowRegistryImpl`, the two `GeneralSettingsView`s).

**3. Shared `WindowViewScaffold`**
Extracts the window frame / header / stream-render loop / scroll handling shared by desktop and mobile into one `commonMain` scaffold, with the toolkit-specific pieces (surface, header, scrollbar, context menu, settings dialog, dialog content) supplied as slots. Desktop/mobile `WindowView` become thin wrappers.

## Testing

- `:compose:compileKotlinJvm`, `:compose:compileAndroidMain`, `:compose:ktlintCheck`, `:core:ktlintCheck` pass.
- The desktop app builds and replays real session logs end-to-end (389-line and 21,619-line) with no errors; against `v3.1.0-beta.7`, replaying the 21,619-line log trends ~15% faster (noisy macro benchmark).

## Notes

- No unit tests cover `ComposeTextStream`, so the buffer rewrite rests on reasoning + the replay runs above.
- `OffsetList` intentionally uses referential equality (documented on the class) — safe only because the sole consumer reads/indexes it and never compares two snapshots by content.

### DIFF
--- a/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/window/ComposeTextStream.kt
+++ b/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/window/ComposeTextStream.kt
@@ -3,6 +3,9 @@ package warlockfe.warlock3.compose.ui.window
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.buildAnnotatedString
 import co.touchlab.kermit.Logger
+import kotlinx.collections.immutable.PersistentList
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.collections.immutable.toPersistentList
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
@@ -45,6 +48,7 @@ class ComposeTextStream(
     private var markLinks: Boolean,
     private var showImages: Boolean,
     private var showTimestamps: Boolean,
+    private var suppressPrompts: Boolean,
     private val highlights: StateFlow<HighlightIndex>,
     private val names: StateFlow<List<ViewHighlight>>,
     private val alterations: StateFlow<List<CompiledAlteration>>,
@@ -53,9 +57,18 @@ class ComposeTextStream(
     private val workQueue: StreamWorkQueue,
     private val scope: CoroutineScope,
 ) : TextStream {
-    private val cacheLines = ArrayList<CachedLine?>(maxLines)
-    private val finishedLines = ArrayList<StreamLine>(maxLines)
+    // ArrayDeque so trimming the oldest line (removeAt(0)) once the buffer fills is O(1) rather than
+    // the O(n) front shift of an ArrayList.
+    private val cacheLines = ArrayDeque<CachedLine?>(maxLines)
+    private val finishedLines = ArrayDeque<StreamLine>(maxLines)
     val lines = MutableStateFlow<List<StreamLine>>(emptyList())
+
+    // The displayed lines are an append-only persistent list plus a logical start index. Once the
+    // buffer fills, evicting the oldest displayed line (which happens on every newline) is then an
+    // O(1) start bump rather than an O(n) front removal, while appends stay O(log n). The backing is
+    // compacted once the dropped prefix grows large, amortizing that to O(1) per line.
+    private var displayBacking: PersistentList<StreamLine> = persistentListOf()
+    private var displayStart: Int = 0
     private val components = mutableMapOf<String, StyledString>()
 
     private var nextSerialNumber = 0L
@@ -125,8 +138,9 @@ class ComposeTextStream(
                     isPrompt = isPrompt,
                 )
             cacheLines[cacheLines.lastIndex] = cachedLine
-            finishedLines[finishedLines.lastIndex] = cachedLineToStreamLine(cachedLine, serialNumber)
-            linesUpdated()
+            val newLine = cachedLineToStreamLine(cachedLine, serialNumber)
+            finishedLines[finishedLines.lastIndex] = newLine
+            replaceLineInView(serialNumber, newLine)
         }
     }
 
@@ -175,7 +189,7 @@ class ComposeTextStream(
         val line = cachedLineToStreamLine(cachedLine, serialNumber)
         finishedLines.add(line)
         line.text?.let { playSound(it.text) }
-        linesUpdated()
+        appendLineToView(line)
     }
 
     private fun addComponentLocations(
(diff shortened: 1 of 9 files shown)


sproctor:
## Code review (post-merge): buffer perf, suppress-prompts, shared window scaffold

**Bottom line:** no confirmed user-visible correctness regression. The buffer rewrite's core invariants (serial-sorted `displayBacking`, binary-search bounds, `displayStart <= size`, eviction loop) all hold under verification. The findings below are one latent correctness hazard, a few efficiency gaps, and some cleanup/altitude items. Ranked most severe first. (Line numbers are against the post-merge `ComposeTextStream.kt`; later PRs have since drifted the window-view files.)

### 1. `OffsetList` breaks the `List` equals/hashCode contract (latent correctness)
`ComposeTextStream.kt:358-369` (init at `:64`)
`equals` is referential (`this === other`) and `hashCode` returns `size`, on an `AbstractList`. This is documented-intentional to make `StateFlow` dedup O(1), and the single current consumer (index-only `collectAsState`) is unaffected. But it is an asymmetric-equals contract violation: `lines` is initialized to `emptyList()` (content equality), so `EmptyList.equals(emptyOffsetList)` is `true` and the first empty publish is suppressed (harmless today: empty -> empty). Any future consumer that diffs two snapshots, uses one as a Set/Map key, or asserts `List` equality in a test will silently misbehave.
**Fix:** initialize the flow with `OffsetList(persistentListOf(), 0)` so every old value shares the referential contract; drop/clarify `hashCode` (see #8).

### 2. `replaceLineInView` "now filtered out" branch never compacts (minor correctness/memory)
`ComposeTextStream.kt:321-325`
The append path calls `compactBacking()` (`:293`); the `removeAt` branch publishes but never compacts. Under name-filter churn where component updates flip displayed lines out of the filter with no intervening newline, the evicted prefix `[0, displayStart)` is retained. Verified: retention is **bounded** by `maxLines` and cleared by the next append or any `linesUpdated()` rebuild, so it is a bounded/temporary inconsistency, not an unbounded leak. (The stronger `displayStart > size` / IndexOutOfBounds claim was refuted; the `backingIndex >= displayStart` guard preserves the invariant.)
**Fix:** call `compactBacking()` after `removeAt` too.

### 3. `updateComponent` emits one `StateFlow` update per affected line (efficiency)
`ComposeTextStream.kt:254-260` -> `updateLine` -> `replaceLineInView` -> `publishLines()`
A component appearing on K buffered lines produces K fresh `OffsetList` instances and K emissions for one value change. Pre-existing (the old per-line `linesUpdated()` also emitted K times), and the refactor made each emission strictly cheaper (O(log n) `set` vs O(n) rebuild), but the over-emission survived. Compose coalesces recompositions per frame, so the real waste is K allocations + K persistent-list copies per update of a hot multi-line component.
**Fix:** add a mutate-only variant (no publish), batch the K `set`s, then `publishLines()` once.

### 4. Image path left on the legacy full-rebuild while text was converted (efficiency/altitude)
`ComposeTextStream.kt:229-245` (`appendResource`)
The PR converted `doAppendLine` to incremental `appendLineToView`, but `appendResource` still calls O(n) `linesUpdated()` and never calls `removeLines()`. Pre-existing, but now inconsistent with the new fast path: each image does a full re-filter/rebuild and resets `displayStart=0`, and an image-only window grows `finishedLines`/`cacheLines` unbounded (no eviction). Text-interleaved windows self-bound on the next text line.
**Fix:** route image lines through `appendLineToView` (they pass the prompt filter trivially) and add `removeLines()`.

### 5. Prompt/line visibility is now split across two layers (altitude)
`ComposeTextStream.kt:433` (`linePassesFilter`) + the composable render guard (`isShowing` / `isPreviousPrompt`)
`suppressPrompts` is added as a second ad-hoc boolean AND'd next to `nameFilter`, while the composable still independently filters `isShowing(openWindows)` and dedups consecutive prompts. Filtering lives in both the data layer and the view. `lines.size` still counts view-hidden rows, which the scroll/sticky logic (`scrollToItem(lastIndex)`, `totalItemsCount-1`) treats as real items (pre-existing). A third filter means another copy-pasted setter + AND clause.
**Fix (deeper):** model filters as a predicate set (`filters.all { it(line) }`) with one "filters changed -> rebuild", and consider routing `openWindows` into the stream so one displayed list is the single source of truth.

### 6. Duplicated relayout-scheduling boilerplate (simplification)
`ComposeTextStream.kt:417-425` (`setSuppressPrompts`), `:451-459` (`setNameFilter`), `:401-407` (`setMaxLines`), `:92-99` (names `onEach`)
Four copies of `... { workQueue.submit { mutex.withLock { linesUpdated() } } }`.
**Fix:** extract `private fun scheduleRelayout()`.

### 7. Per-setting fan-out boilerplate in the registry (altitude)
`WindowRegistryImpl.kt:91-100`
Each per-stream setting (now `suppressPrompts`) duplicates observe + `onEach { streams.forEach { ... } }` + `stateIn`, plus a matching setter and a constructor arg - four edit sites that must stay in lockstep (new streams get the value via the constructor, existing ones via `onEach`; easy to wire one and forget the other).
**Fix:** collect per-stream display settings into one observable object the stream subscribes to.

### 8. `OffsetList.hashCode()` is dead and misleading (simplification)
`ComposeTextStream.kt:368`
`hashCode` returns `size` but the list is never used as a hash key or in content comparison; paired with referential `equals` it implies content hashing that does not exist.
**Fix:** remove the override (or return `identityHashCode`) and document `equals` as identity-only. Pairs with #1.

### 9. Fully-qualified `Box` instead of an import (simplification nit)
`DesktopWindowView.kt` (surface slot uses `androidx.compose.foundation.layout.Box(...)`; the `import` was removed in the diff)
Inconsistent with the rest of the file; signals a missing import rather than intentional disambiguation.
**Fix:** re-add `import androidx.compose.foundation.layout.Box` and use the plain name.

### 10. Binary search runs for filtered-out partial lines that can never be found (minor efficiency)
`ComposeTextStream.kt:307-313`
When `suppressPrompts`/`nameFilter` excludes a streaming partial line, each `doAppendPartial` update runs `binarySearch(displayStart, size)` that necessarily misses, then hits the no-op branch - O(log n) per partial append for zero visible effect.
**Fix:** short-circuit when `!shouldShow` and the line is not the tail.

---

**Checked and cleared (not bugs):** the `displayStart`/`size` invariant and `OffsetList` bounds (no IndexOutOfBounds); the cross-thread `suppressPrompts` read (executes under the mutex on the work queue and is reconciled by the queued rebuild - matches, and is safer than, the existing `setMarkLinks`/`setShowImages` convention); the scaffold slot delegation (all slots forwarded with correct order/values; `addTextContextMenuOptions` moved cleanly, not duplicated); the constructor change (single named call site); and conventions (`kotlinx-collections-immutable` is already in the version catalog; no non-ASCII introduced). The `doAppendPartial` empty-buffer `last()` crash is real but pre-existing and byte-identical to before this PR (requires `maxLines`=1 mid-partial), so it is out of scope as a regression.

<sub>Generated with Claude Code `/code-review`</sub>

PR #227, opened 2026-07-07
Add global font color and per-messagetype size

raidekoptix:
I scribbled together a fix to restore font color and per-message sizing functionality lost in the last font settings overhaul. This PR is just to demonstrate what I have in mind. The color for the default font is a bit redundant with the "Default" message type further down on the Appearance tab; it can probably be removed. Also swapped Bold/Italic/Underline checkboxes with buttons to make room for size box.


sproctor:
this was addressed by #232

PR #257, opened 2026-08-02
Implement the Friends & Enemies (befriend) panel

sproctor:
Implements the server-driven befriend panel ("Friends && Enemies" in GS4/DR), verified against the real Wrayth client in the wrayth lab and end to end by running the desktop app against the lab's fake server with payloads captured from real game logs.

## Protocol

- New `menulink`/`menuimage` handlers producing `PanelObject.MenuLink`/`MenuImage` (value/name, exist, noun). Clicking one requests the server context menu (`_menu #<exist> <id>`) through the same machinery as `<a exist=>` text links; the response pops up anchored at the widget, with cmdlist labels/commands resolved per noun/exist.
- Widget `cmd` attributes are now sent verbatim and without the normal command echo, and the `echo` attribute is displayed instead when present. This matches Wrayth: real payloads pair internal commands (`cmd='_ready weapon'`) with readable echoes (`echo='ready weapon'`), which previously echoed raw.
- `_DBOPEN`/`_DBCLOSE` are sent when the user shows/hides a dialog panel, so the server knows whether that panel's updates are worth sending (it answers `_DBOPEN` by re-sending the contents).
- `openDialog` titles collapse Windows menu-mnemonic `&&`, so the panel shows "Friends & Enemies" like the real client.

## Faces

The demeanor faces are skin data, not a code table: new `compose/skin/skin.json` entries map the wire names to bundled Material sentiment icons with per-theme demeanor tints (friendly=gold, warm=green, neutral=gray, reserved=pale blue, cold=blue, crossFace=red X — matching Wrayth's demeanor colors, using hues the skin already uses). `PanelImage` resolves a skin entry's drawable `name` the same way the indicators do, so a ported skin carrying real face bitmaps still wins.

The reserved face is `sentiment_content` (lidded eyes, flat mouth): the demeanor scale needs it colder than neutral, and `sentiment_calm`'s slight smile would read warmer.

## Testing

- 4 new protocol tests built from real log payloads (menuLink/menuImage rows, the full befriend line, the `&&` title).
- Live run against the lab server: remove button echoes `>befriend clear 1` and sends it; name click sends `_menu #-10246518 0`; a served menu response pops up with substituted labels and a category submenu, and clicking an item sends `befriend clear #-10246518`; hide/show sends `_DBCLOSE`/`_DBOPEN`; a re-sent `dialogData` repopulates the reopened panel.
- `check -PiosSkip=true -PlintSkip=true`, `jvmTest`, and ktlint all pass.


sproctor:
Pushed a fix for the two regressions reported against this branch: hidden panels reopening on every login, and duplicate panels when a window was already open before its openDialog arrived.

Root cause (predates this branch, from #252, but surfaced while testing it): GS4 announces nearly all panels **before** the `<app>` tag, so `openWindowFromGame` sampled a still-null character id and silently skipped both the `isHidden` and saved-location guards. Hidden panels reopened (the flag was persisted but never consulted — no `openWindow` saves appear in the session log because the null id also skipped the save), and saved panels opened at their protocol location, after which the layout restore appended a second copy. DR sends `<app>` before its dialogs, which is why the DR-shaped replay passed.

Fix: wait for the character id before evaluating the guards (10s timeout for a server that never identifies one), and make the layout restore skip windows already in the layout — which also covers re-restores on reconnect.

Verified by replaying a captured GS4 login through the wrayth lab against a character seeded with a hidden `injuries` panel and a `bounty` saved at TOP while announced at RIGHT: before the fix, injuries reopened and bounty appeared twice; after, injuries stays hidden, bounty opens once at its saved dock, and every game-announced panel saves under the real character id.

PR #259, opened 2026-08-03
Keep the window order stable across launches

sproctor:
## Problem

A user reported that with the latest version, every subsequent launch/connect re-arranges the order of their dock windows.

Two writers were corrupting the saved positions:

- **Game-announced panels raced the layout restore.** An `openDialog` handler and `restoreWindowLayoutOnConnect` both wake on the same `characterId` emission, so a panel with no saved dock recorded a position computed from a dock list the restore had not populated yet, colliding with the saved windows' positions. beta.26 (#252) made this much worse by evaluating the saved-location guards against a still-null character id (every resident panel re-opened and overwrote its saved placement on every login); beta.27 (#257) fixed the guards but the position race remained.
- **`openWindowAt` read the dock list's `lastIndex` inside the launched coroutine**, after later opens had already appended, so back-to-back opens into the same dock recorded the same position.

Duplicated positions make the restore's `ORDER BY position` resolve ties by rowid - an order that need not match what the user last saw - and the DAO's gap arithmetic assumes unique positions, so dragging windows around on top of duplicates corrupted them further. The order kept shuffling on every connect until a full-dock drag happened to rewrite every row.

## Fix

- Game-announced opens wait for the layout restore to finish (a `layoutRestored` flag set by the restore in a `finally`, with a timeout covering a stuck restore), so they append after the saved layout instead of racing into it.
- `openWindowAt` captures the position at the moment the window is appended to the dock list.
- The restore renumbers each dock's saved positions to 0..n before reading them (`WindowSettingsDao.normalizePositions`, a transaction), healing the duplicates and gaps existing databases already carry. Windows out of the layout are untouched and a clean layout writes nothing.

## Caveat

The renumbering makes the order stable but cannot recover what beta.26 overwrote: a panel it moved keeps its healed spot, so affected users may need to arrange their docks one last time - after that the layout sticks.

## Testing

- New `WindowPositionNormalizationTest` covers duplicate renumbering in sort order, gap compaction with null positions placed first, per-dock independence, out-of-layout rows being left alone, and no writes on a clean layout.
- `./gradlew jvmTest` passes; core and compose also compile for Android.


sproctor:
## Code review findings

Multi-agent review of c5ca1c7e: 10 finder angles, candidates deduplicated and adversarially verified one by one; 2 candidates were refuted and excluded (null-position fronting, MAIN-dock guard bypass). 15 findings survived. **Findings 1-4 are the ones to act on before shipping**: they are the residual writers and failure paths that can still reproduce, or newly cause, the cross-launch shuffle and lost windows this PR exists to fix.

### Fix before merge

**1. Transient panels still skew persisted dock positions** - `GameViewModel.kt:1355` (CONFIRMED)

Persisted positions are still computed from UI dock lists that include transient (`resident=false`) panels which have no DB row (`openWindowAt` appends to the list, then returns at `canSaveWindow` before persisting; `removeWindow` also returns before any `closeGap`), so purely game-driven flows keep writing duplicate positions after this fix. Note `resident` defaults to false when the openDialog attribute is absent.

*Failure:* Restore docks A(pos 0); game opens transient T (UI [A,T], no DB row); resident B opens -> newPosition=2 persisted (DB: A=0, B=2); T closes (DB untouched); resident C opens -> newPosition=2 persisted, colliding with B. Next connect `normalizePositions` freezes the tie in arbitrary order (empirically name order, not the on-screen order), reproducing the exact cross-launch shuffle this commit set out to eliminate. Deeper fix: assign position atomically in SQL (`COALESCE(MAX(position)+1,0)` per dock) or count only savable windows.

**2. Restore has no catch; a normalize failure drops all saved windows** - `GameViewModel.kt:554`

The restore's `try` block has `finally` but no `catch`, and the commit adds `normalizePositions` (a write transaction) as a new hard prerequisite: any exception from it (or the settings read) propagates out of `onEach`, permanently cancels the characterId collector, and reaches viewModelScope unhandled (app crash on Android), while the `getWindowLocation` guard then silently drops every window with a saved dock for the whole session.

*Failure:* A second instance or backup tool holds the prefs DB write lock (or disk full) when the user connects; `normalizePositions` throws -> no saved windows are restored and none of them ever open (their openDialog is swallowed by the saved-location guard), with no retry and no error shown. Pre-commit this path was read-only and could not fail this way. Needs a catch that logs and falls back to restoring from the un-normalized read.

**3. closeDialog bypasses the gate; a parked open inverts a close** - `GameViewModel.kt:1333` (CONFIRMED)

The `layoutRestored` gate parks only game opens while closeDialog handling runs immediately in the event collector, so openDialog-then-closeDialog for the same panel inverts: the close removes nothing (window not docked yet) and nulls the DB row, then the parked open resumes, passes both guards, opens the panel, persists it, and clears hidden. The second wait widens the pre-existing inversion window to include the whole restore.

*Failure:* Game sends openDialog X then closeDialog X while the open is parked on the gate: X ends the session open on screen and saved as open in the DB even though the server's last instruction was close, and it comes back on every future connect. Fix: route closes through the same wait, or drop a parked open when a later close for the same name has been seen.

**4. `finally` releases the gate even after a failed restore** - `GameViewModel.kt:587`

`finally` sets `layoutRestored=true` even when the restore threw partway or before adding anything, so gated game opens proceed against an empty or half-populated dock and persist positions computed from the deficient list, writing fresh duplicates.

*Failure:* Saved RIGHT dock [thoughts@0, logons@1]; the restore fails after adding only thoughts; the gate releases; game-announced befriend (no saved dock, so both guards pass) appends at UI index 1 and persists position=1, colliding with logons' saved position 1, and nothing re-heals until the next connect. A separate failed-restore flag (or not marking restored on exception) would keep the fallback conservative.

### Remaining findings

<details>
<summary>11 more: narrower races, pre-existing writers, test fidelity, cleanup</summary>

**5. Tie-order comment is false and the test pins an order production reverses** - `WindowSettingsDao.kt:235`

The doc comment's healing mechanism ("duplicated positions make ORDER BY resolve ties by rowid") is false and the new test certifies the wrong order: empirically the (characterId,name) PK autoindex plan returns ties in NAME order (befriend before thoughts, the reverse of the test's asserted heal), and `@Insert(onConflict=REPLACE)` reassigns rowids anyway; SQLite guarantees no tie order at all.

*Failure:* Run the test fixture's rows through the real DAO: production heals to [befriend:0, thoughts:1, logons:2] while `duplicatePositionsAreRenumberedInSortOrder` (WindowPositionNormalizationTest.kt:32) pins [thoughts:0, befriend:1, logons:2], so the test green-lights an order production reverses and the comment teaches a nonexistent mechanism. Fix the comments to say the tie order is unspecified and assert renumbering invariants (0..n, no duplicates) instead of exact tie order.

**6. User-initiated opens skip the layoutRestored gate** - `GameViewModel.kt:1282`

User-initiated opens (desktop window menu DesktopGameView.kt:175, LargeGameView.kt:127, settings toggles via WindowSettingsLiveContext) call `openWindowAt` with no `layoutRestored` wait, so the duplicate-position race this PR fixes for game-announced opens survives on every user-facing path.

*Failure:* A click lands between the characterId emission and the restore's completion (milliseconds normally, but up to the full restore duration when the DB is slow): `openWindowAt` computes newPosition=0 from the not-yet-populated TOP dock and persists it while the saved layout also holds a TOP window at 0; next launch tie-breaks in an order the user never saw. Gating `openWindowAt` itself (where the position is computed) covers all callers.

**7. Drag reorder persists as unserialized per-row updates** - `GameViewModel.kt:1276`

`changeWindowPositions` persists a drag-reorder as one independent `setPosition` transaction per row, reading names from the live list across suspension points with no transaction or mutex, leaving it an unserialized writer of the same rows `normalizePositions` heals only at connect.

*Failure:* Two quick same-dock drags, or one drag interleaving with `moveWindowToPosition`'s gap arithmetic or a Room writer-thread reordering, land a stale (name,index) UPDATE last: two rows share a position for the rest of the session and the next connect freezes them in arbitrary tie order, silently discarding the arrangement the user just dragged into. A single `@Transaction` DAO method taking the full ordered name list would make the write atomic.

**8. Settings toggle writes a cross-dock count as a TOP position** - `DesktopWindowsSettingsSection.kt:149`

The settings screens write `repository.openWindow(characterId, name, TOP, position = openWindows.size)` where `openWindows` counts saved windows across ALL docks (same in WindowsSettingsSection.kt:144), writing gapped and even colliding TOP positions straight to the DB with no gate, immediately re-breaking the 0..n invariant `normalizePositions` just established.

*Failure:* Character has 2 TOP + 3 RIGHT windows: toggling a window on writes TOP position 5; open B (position 5), close another TOP window (closeGap shifts B to 4), then open C -> `openWindows.size` is 4, colliding with B. The duplicate sorts arbitrarily until the next connect and can flip the visible order. Count `getByLocation(TOP).size` (the DAO's own per-dock idiom) instead.

**9. Gate timeout silently falls back to the old race** - `GameViewModel.kt:1333`

The null result of `withTimeoutOrNull(10_000)` on `layoutRestored` is discarded with no logging, so a slow-but-alive restore is silently raced exactly as before the fix, the two sequential 10s waits can stack to 20s, and a config store that never emits makes `observeWindowSettings().first()` hang so `finally` never runs and every panel that session eats the full 10s before opening.

*Failure:* Cold start with DB/WAL recovery pushes normalize + first() past 10s during the login burst: all parked `openWindowFromGame` coroutines time out together and append mid-restore, recording positions that duplicate saved ones, with nothing in the logs to explain the next launch's reshuffle. At minimum log the timeout; ideally re-check the dock lists after the restore completes.

**10. Stale newState/newPosition on CAS retry in openWindowAt** - `GameViewModel.kt:1354`

`newState` and `newPosition` are assigned inside the CAS-retried `update{}` lambda and never reset at lambda entry, so a retry that lands in the already-exists branch keeps stale values from the failed attempt and the code below persists `openWindow` for a window this call did not add; latent only because every mutator of the four dock flows currently runs on Main.immediate.

*Failure:* Any future off-main mutator (or dispatcher change) makes `update` contend: attempt 1 sets newState/newPosition, CAS fails because a concurrent call added the same name, attempt 2 returns states unchanged, yet `val state = newState ?: return` passes and a second openWindow write lands with a stale position, plus `buildWindowUiState` ran twice. Reset both vars at the top of the lambda (or derive added-ness from the returned snapshot).

**11. Restore snapshot races closeDialog for saved windows** - `GameViewModel.kt:555`

The restore's one-shot settings snapshot is not serialized against game closes: a closeDialog for a saved window landing across the normalize/first() suspension points removes nothing from the not-yet-populated dock while its async DB write nulls the row, and the loop then docks the window from the stale snapshot anyway (pre-existing shape, rewritten by this commit).

*Failure:* Game sends closeDialog W in the milliseconds the restore is suspended: W appears docked on screen with no DB row behind it, the dock's UI count exceeds the DB count so the next gated open records a gapped/colliding position, and W silently vanishes at the next connect. Narrow and mostly self-healing, but the restore block now owns ordering and should account for closes observed since the snapshot.

**12. openWindowAt presence check scans one dock only** - `GameViewModel.kt:1351`

`openWindowAt`'s presence check scans only the destination dock while `openWindowFromGame` checks all four lists, so a call for a window already docked elsewhere would append a second WindowUiState sharing the same stream and rewrite the saved row to TOP; currently defended only by the callers' toggle UIs, which route open windows to closeWindow.

*Failure:* A caller invokes `openWindow(name)` for a LEFT-docked window during a UI desync (openWindows collectAsState still on its initial empty list, or a window opened while characterId was null and thus never persisted): two copies render until restart, then only the TOP copy restores and the layout visibly changes across launches. Move the all-lists check into `openWindowAt`.

**13. Reconnect comment describes an unreachable situation** - `GameViewModel.kt:557`

The rewritten guard comment ("A reconnect runs the restore again over a populated layout") describes a provably unreachable situation: characterId is written exactly once per client (WraythClient.kt:361, never nulled) and reconnect() builds a fresh client and GameViewModel, so the guard's real live purposes (the ungated user-open race and the 10s timeout fallback) go undocumented.

*Failure:* A future reader deletes the apparently redundant guard because its stated rationale is dead, and the user-open and timeout races then produce duplicate WindowUiStates in the dock lists; conversely, if character re-identification is ever added, `layoutRestored` is already true and the dock lists never reset, so the comment's scenario would corrupt positions rather than be handled.

**14. Test fake duplicates InMemoryWindowSettingsDao** - `WindowPositionNormalizationTest.kt:115`

The new ~115-line private InMemoryDao is a near-verbatim second copy of the private InMemoryWindowSettingsDao in WindowHiddenStateTest.kt:120 (identical openWindow/doCloseWindow/getByName bodies plus another block of error("unused") stubs), and its entity() helper is the third and fourth copy of the all-defaults WindowSettingsEntity fixture; one shared jvmTest fake plus a windowSettingsEntity() factory would serve both tests.

*Failure:* Every WindowSettingsDao signature or entity change must now be repeated across two ~100-line fakes and four construction sites in sibling test files, and the fakes have already diverged (one models the query's NULLs-first ordering, the other error()s getByCharacter), so future tests can silently exercise different DB semantics.

**15. layoutRestored latch would be safer as CompletableDeferred** - `GameViewModel.kt:415`

A one-shot, never-reset latch is modeled as MutableStateFlow(false) plus `finally { value = true }` plus `first { it }`; `CompletableDeferred<Unit>` (complete() in the finally, await() at the wait site) encodes the one-shot stays-set semantics in the type instead of in a three-line comment.

*Failure:* The boolean-flow shape invites a future "fix" that resets it to false on reconnect (a StateFlow looks resettable, a completed Deferred does not), which would stall every game-announced open after a reconnect on the 10s timeout before it appears.

</details>

sproctor:
All 15 review findings are addressed in cd46e007. The structural change that resolves most of them: a window's persisted position is now computed inside the database, never derived from an on-screen list index.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Transient panels skew persisted positions | `WindowSettingsDao.openWindowAtEnd` computes the position in a transaction from the dock's saved rows (`getByLocation` max + 1), so on-screen transients and concurrent opens can't skew it |
| 2 | Restore has no catch; normalize failure drops all windows | Restore body wrapped in catch/log (CancellationException rethrown); a `normalizePositions` failure has its own catch and falls back to restoring the un-normalized read |
| 3 | closeDialog bypasses gate; parked open inverts a close | `closeWindowFromGame` waits on the same gate, and both handlers carry a per-window instruction stamp - whichever is no longer the game's latest instruction for that window drops itself, in either direction |
| 4 | `finally` releases gate after failed restore | Still released, deliberately - now safe because parked opens append at the DB end of their dock rather than persisting a UI-derived index; holding the gate would stall every panel on its timeout. Rationale documented at the release site |
| 5 | Tie-order comment false; test pins order production reverses | All rowid claims replaced with "SQLite leaves the order of duplicated positions unspecified"; the test now asserts renumbering invariants (0..n distinct, untied rows keep their spot) instead of a tie order |
| 6 | User-initiated opens skip the gate | Collision impossible now that the position is DB-computed; the user path goes through the same `openWindowAtEnd` |
| 7 | Drag reorder persists as unserialized per-row updates | `setPositions` writes the dock's full order (captured before suspending) in one `@Transaction`; also repairs any stale positions the dock carried |
| 8 | Settings toggle writes cross-dock count as TOP position | Both settings screens now call `openWindowAtEnd` |
| 9 | Gate timeout silently falls back to the old race | Timeouts logged in both the open and close paths |
| 10 | Stale newState/newPosition on CAS retry | `newState` reset at lambda entry; `newPosition` no longer exists |
| 11 | Restore snapshot races closeDialog | Covered by the close path waiting on the gate (see 3); a close arriving mid-restore acts after the dock is populated |
| 12 | Presence check scans one dock only | The all-docks check moved into `openWindowAt`, covering every caller |
| 13 | Reconnect comment describes unreachable situation | Comment now names the guard's live purposes: user opens and timed-out game opens landing before the restore |
| 14 | Duplicated test fake | One shared `InMemoryWindowSettingsDao` (plus `windowSettingsEntity()` factory and `dock()` helper) now serves both test classes; the private copies are gone |
| 15 | Latch should be CompletableDeferred | `layoutRestored` is a `CompletableDeferred<Unit>` |

Also removed as fallout: `WindowSettingsRepository.openWindow` and `setPosition` (no callers remain - placement goes through `openWindowAtEnd`, reorders through `setPositions`).

Verified with `./gradlew jvmTest` (includes the renamed `WindowPositionsDaoTest`, now also covering `openWindowAtEnd` and `setPositions`) plus Android compilation of core and compose.

PR #260, opened 2026-08-04
Remember a closed window's placement and reopen it there

sproctor:
## What

A closed window now remembers where it was and goes back there when it reopens - whichever path reopens it (the game's `openDialog` for a panel it closed with `closeDialog`, the window list, or a settings toggle, online or offline). Previously closing nulled the saved dock and position, so a game-closed panel drifted to the end of its dock on the next login and a user-reopened window always landed in TOP.

Two commits:

**Remember a closed window's placement (255ca016)**
- Schema v21: window rows keep `location`/`position` when closed, and a new `open` column (backfilled from `location IS NOT NULL`; manual migration because of the backfill) marks layout membership. The restore, the settings screens, and the openDialog guards now key off `open`.
- New `reopenWindow` DAO transaction puts a window back at its remembered spot and clears the hidden flag; windows never placed fall back to what each path did before (protocol location for game opens, TOP for user opens).
- Exports carry the flag with backward compatibility (older exports read placed-in-layout as `location != null` on import). The dead `moveWindow` DAO helper is removed.

**Preserve order, not just an index (66fa0f5d)**
- A position is a slot in the dock's total order, open and closed windows as one sequence. Close and reopen are flag flips that keep the slot, so a window returns between the same neighbors and windows closed together reopen in their original relative order regardless of reopen order. (The first cut froze an index and clamped it on reopen, which could invert the order of windows closed together.)
- The operations that change order maintain the sequence instead: appends go after every slot (no collisions with remembered ones), a dock reorder merges the new open order around closed windows' kept slots, cross-dock moves shift closed slots with their neighbors, and the connect-time heal renumbers each dock's whole sequence 0..n - still killing the duplicate positions the previous PR (#259) set out to heal, now without disturbing remembered slots.
- `reopenWindow` returns the window's rank among open windows, which is what the on-screen dock list needs for inserting (transient panels on screen don't shift where it lands).
- Simplification fallout: the `doCloseWindow`/`closeWindow` split collapses into one query and `closeGap` is gone entirely.

## Behavior note

Closes that happened before upgrading have nothing remembered (their rows were already nulled), so those windows reopen at the defaults once; every close from v21 on remembers.

## Testing

- New DAO tests: reopen at the remembered spot, windows closed together reopening in their original order, slot preservation through the connect-time heal, reorder keeping a closed window's slot from the top, appends landing past remembered slots, and never-placed fallback. Hidden-flag tests cover reopen clearing hidden.
- `./gradlew jvmTest` passes; core and compose compile for Android; generated Room schema `21.json` matches the migration's `ALTER TABLE` exactly.


sproctor:
## Code review findings

Multi-agent review of the two commits on this branch (255ca016, 66fa0f5d): 8 finder angles plus a line-by-line pass, candidates adversarially verified, plus a gap sweep. Verifiers CONFIRMED the MAIN crash, the open/close race, and the drag-index mismatch; the login-gate gap was downgraded to plausible; two candidates were refuted (a generic setPositions interleaving race - the source dock's own transaction repairs it - and a null-position-reopen claim). 15 findings survived.

### Fix before merge

**1. Reopening a remembered `MAIN` location crashes** - `GameViewModel.kt:1329` (CONFIRMED)

`placeWindowUi` feeds the DB location into `getWindowUiStatesForLocation`, whose else branch is `error("Change position error: Invalid window location")`. Legacy DBs contain `location=MAIN` rows (2022-era code wrote `defaultLocations mapOf("main" to WindowLocation.MAIN)`), MIGRATION_20_21 backfills them `open=1`, and the sidebar window list includes "main" (residentWindows has no main exclusion). Hide-then-show on that entry: `openWindow` -> `reopenWindow` returns `WindowPlacement(MAIN, 0)` -> IllegalStateException inside a launched coroutine with no catch (a crash on Android). The old code never fed the DB location back into UI placement (it hardcoded TOP). Same unguarded call in `openWindowFromGame`, which filters only the protocol location, not a saved MAIN.

**2. Legacy `MAIN` row can replace the main text window** - `DesktopWindowsSettingsSection.kt:150`

Same root cause, worse outcome on the disconnected settings path: the Show toggle's `reopenWindow` happily resurrects a legacy `location=MAIN` row as open, and the next connect's restore hits `WindowLocation.MAIN -> _mainWindowUiState.value = uiState`, replacing the main text window with the panel. Same code in WindowsSettingsSection.kt:145. The clean fix covers both findings: `reopenWindow` should treat a remembered MAIN as never-placed, so every caller falls back to its default.

**3. `openWindow` went async; a close can interleave and UI/DB diverge** - `GameViewModel.kt:1318` (CONFIRMED)

The click now suspends through a Room `@Transaction` plus a TOML write (setHidden under a file lock) before the window docks, and nothing covers user opens (the stamp map is game-path only). A game `closeDialog` landing in that window removes nothing from the dock lists (window not there yet), writes closed+hidden, then the open resumes and docks it: visible on screen, closed+hidden in the DB, and the game can never reopen it past the isHidden guard. The sidebar dot also flips as soon as `markOpen` commits - before the window is docked - so the next click routes to `closeWindow`. For panels the server receives `_DBCLOSE` then `_DBOPEN` inverted. Master was immune: the dock add happened synchronously at click time. Suggested shape (also fixes the added click latency, finding 14): place the UI first from the already-observed `windowSettings` flow, persist afterward.

**4. Cross-dock drag passes a UI index where the DAO now expects a rank** - `GameViewModel.kt:1232` (CONFIRMED)

`DragDropState.computeDropTarget` derives insertionIndex from every on-screen window, transient panels included, but the rewritten DAO `moveWindowToPosition` interprets it as a rank among the dock's open saved rows. With a transient docked, the UI inserts in one place and the DB persists another; the next connect restores the wrong order. The diff added exactly the needed rank-to-index mapping in `placeWindowUi` but not the inverse here; both drop handlers (DesktopGameView.kt:273, GameView.kt:546) violate the DAO's documented contract.

### Remaining findings

<details>
<summary>11 more: login-window races, rank mapping, export compat, migration test gap, cleanup</summary>

**5. User opens skip the login gates** - `GameViewModel.kt:1323` (PLAUSIBLE)

`openWindow` awaits neither characterId nor layoutRestored (unlike `openWindowFromGame`). The sidebar is populated before characterId arrives and renders every entry as closed while windowSettings is still empty, so a click during login maps its DB rank onto a not-yet-populated dock and pins the window at slot 0; nothing ever re-sorts a dock afterwards, and a later drag persists the wrong order. Sibling case with no race: a click before characterId is set takes the TOP fallback and skips the DB write entirely, stranding the window in TOP while the DB says elsewhere.

**6. placeWindowUi's rank denominator differs from the DB's** - `GameViewModel.kt:1454`

The DB rank counts open rows; the UI mapping counts on-screen windows passing `canSaveWindow`, which is also true for windows with no row yet (unannounced, or an `openWindowAtEnd` write still in flight). When the sets differ, the reopened window inserts on the wrong side of a neighbor. The mapping needs the rank among windows with saved open rows, not among savable ones.

**7. moveWindowToPosition's elvis conflates null slot with append** - `WindowSettingsDao.kt:195`

`getOrNull(index)?.position ?: append` treats a target row with a NULL position (a real shape: imports can write them, and they sort first) the same as index-past-the-end, silently appending to the bottom instead of inserting at the top. Narrow in practice since the connect heal compacts NULLs, but a mid-session import re-creates them.

**8. New exports resurrect closed windows in older builds** - `ExportRepository.kt:244`

New exports write location/position for closed windows plus `open=false`, but older builds parse with ignoreUnknownKeys and treat `location != null` as docked: importing brings every closed-but-remembered window back docked, combined with hidden=true - a state the old build cannot otherwise produce. (Old-export-into-new-build is handled correctly.)

**9. Non-resident row drifts to the dock end on every reorder** - `WindowSettingsDao.kt:304`

A window that got an open row before the game announced it `resident=false` drops out of `names` forever while its row stays open; `setPositions` then appends it after every named window on each reorder and the restore keeps docking it at the bottom, drifting the saved order from the screen.

**10. MIGRATION_20_21 has no real-SQLite test** - `Migrations.kt:32`

The backfill is exercised only through the in-memory fake; no test in the repo runs any Room migration. A defect would fail at first launch for every upgrading user, and openVersionedDatabase then deletes the failed target and re-seeds on every launch - a permanent crash loop. The repo has shipped exactly this class of ships-green-crashes-on-launch escape before (v3.1.0-beta.21). A migration test seeding a v20 fixture with placed/closed/style-only rows would close it.

**11. Reopen-or-default policy duplicated at four sites** - `GameViewModel.kt:1394`

openWindow, openWindowFromGame, and both settings sections each carry the same reopenWindow-else-fallback block; the MAIN hole already shows the cost (each site needs the same guard). One GameViewModel helper plus a repository-level reopenOrOpenAtEnd would give the invariant one home.

**12. Append-slot rule duplicated** - `WindowSettingsDao.kt:196`

moveWindowToPosition's past-the-end fallback re-implements openWindowAtEnd's max-position+1 computation inline; the two copies can drift.

**13. setPositions has an unreachable arm** - `WindowSettingsDao.kt:308`

The `?: continue` branch cannot fire (closed indices and open names partition the dock exactly), and the closedAtIndex map plus ArrayDeque walk can be a direct merge over the dock rows.

**14. The open click now blocks on DB+TOML before docking** - `GameViewModel.kt:1324`

The old path docked the window in the same frame as the click; the new one waits out a Room transaction and a file-locked TOML write. Placing the UI first from the observed settings flow and persisting afterward restores the synchronous feel and removes the suspension window behind finding 3.

**15. Int.MAX_VALUE as an append sentinel** - `GameViewModel.kt:1412`

`openWindowAt` encodes append as `position = Int.MAX_VALUE` flowing into getOrNull; a nullable Int with null-means-append states the intent and avoids the arithmetic trap.

</details>

sproctor:
All 15 review findings are addressed in 5149af26.

| # | Finding | Resolution |
|---|---------|------------|
| 1 | Reopening a remembered MAIN crashes | `reopenWindow` treats a remembered MAIN as never placed (guard in the DAO transaction, so every caller is covered); DAO test added |
| 2 | Settings toggle resurrects legacy MAIN row over main window | Same guard; the toggle now falls back to appending in TOP |
| 3 | openWindow async; close can interleave, UI/DB diverge | `openWindow` docks synchronously again: placement read from the observed settings (`rememberedPlacement`), UI updated at click time, DB write follows. No suspension between click and dock |
| 4 | Cross-dock drag passes UI index where DAO expects rank | The drop handler translates the on-screen insertion index to the rank among open saved rows before persisting |
| 5 | User opens skip login gates | A click during login defers until the layout restore completes (bounded), so the remembered rank cannot map onto an unpopulated dock; post-login clicks stay fully synchronous |
| 6 | placeWindowUi rank denominator mismatch | The rank now maps over on-screen windows that have open saved rows (from the observed settings), not merely savable ones |
| 7 | moveWindowToPosition elvis conflates NULL slot with append | Split: a target row with a NULL slot means insert at the top; only a rank past the end appends |
| 8 | New exports resurrect closed windows in older builds | Closed windows' placement moved to new `rememberedLocation`/`rememberedPosition` fields; the legacy `location`/`position` are only written while docked, so old builds see closed windows as unplaced. Old-export import unchanged |
| 9 | Non-resident row drifts to dock end on every reorder | `setPositions` keeps any row it wasn't told about (closed or open) in its slot from the top instead of appending, and window-info updates clear the saved row of a non-resident window |
| 10 | Migration 20->21 has no real-SQLite test | New `PrefsMigrationTest` builds a genuine v20 database from the exported schema JSON (all tables, indices, Room identity row, user_version), seeds placed/style-only rows, opens it through Room with `MIGRATION_20_21`, and asserts the backfill; `sqlite-bundled` added to core jvmTest |
| 11 | Reopen-or-default duplicated at four sites | One home per layer: `reopenOrOpenAtEnd` in the repository (both settings screens), `rememberedPlacement` + `placeWindowUi` in the view model |
| 12 | Append-slot rule duplicated | Extracted `appendSlot`, shared by `openWindowAtEnd` and `moveWindowToPosition` |
| 13 | setPositions unreachable arm | Rewritten as a direct walk of the dock sequence (each reorderable slot takes the next name; everything else keeps its slot) - no map, no unreachable branch |
| 14 | Open click blocked on DB+TOML before docking | Same fix as 3: UI first, persistence after |
| 15 | Int.MAX_VALUE append sentinel | `placeWindowUi` takes `position: Int? = null`, null means append |

Verified with `./gradlew jvmTest` (including the new migration test against real SQLite) plus Android compilation of core and compose.

PR #261, opened 2026-08-06
Resolve the dragged window by name when reordering a dock

sproctor:
## Problem

Fixes crash DESKTOP-3J (`IndexOutOfBoundsException: Index: 2, Size: 1` in `changeWindowPositions`, fatal on desktop, seen in production on 3.1.0-beta.27).

A same-dock drag crashed when the dock changed under the gesture. The drop's indices are snapshots: `sourceIndex` is frozen at drag start, and `insertionIndex` comes from item bounds cached at the dock's last recomposition - while the dock list is live. The game closing a panel mid-drag (a transient's `closeDialog`) shrinks the list, and the unguarded `removeAt`/`add` then throw inside the drag-gesture coroutine, which has no handler, taking the whole app down. The crash math matched exactly: the drag layer believed the dock had 3 items while the live list had 2.

There was also a quieter sibling: a stale-but-still-in-bounds `sourceIndex` would reorder whatever window now sat at that index instead of the one the user dragged.

## Fix

- `changeWindowPositions` takes the dragged window's **name** (which `DropResult` already carried), resolves its current index inside the `update {}` - dropping the gesture when the window is gone - and clamps the insertion index to the shrunken list, mirroring the clamp `moveWindowToPosition` already had.
- Both drop handlers (desktop and mobile) pass `result.name`.
- The stale-snapshot trap is removed from the API: `DropResult.sourceIndex`, `DragDropState.sourceIndex`, and `startDrag`'s index parameter are deleted so nothing can consume a drag-start index snapshot again.

## Testing

`./gradlew jvmTest` passes and compose compiles for Android. The crash path is gesture-timing dependent (drop landing in the same frame a window leaves the dock), so it has no unit harness; the fix is defensive resolution against the live list at the single choke point.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved window drag-and-drop reordering for more reliable positioning.
  - Safely handles missing windows and prevents unnecessary layout updates.
  - Correctly adjusts destination positions when moving windows within the same location.

### DIFF
--- a/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/game/GameViewModel.kt
+++ b/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/game/GameViewModel.kt
@@ -1309,15 +1309,24 @@ class GameViewModel(
 
     fun changeWindowPositions(
         location: WindowLocation,
-        fromIndex: Int,
+        name: String,
         toIndex: Int,
     ) {
         val windowUiStates = getWindowUiStatesForLocation(location)
-        var reordered: List<WindowUiState> = emptyList()
+        var reordered: List<WindowUiState>? = null
         windowUiStates.update { states ->
+            // Reset on entry: update{} may retry.
+            reordered = null
+            // The gesture's indices are snapshots, and the dock can change under a drag (the
+            // game closing a panel mid-drag is the everyday case): resolve the dragged window
+            // by name against the current list and clamp the drop index. A stale index crashed
+            // here on a shrunken dock, and a stale-but-in-bounds one would have moved whatever
+            // window sits at it now.
+            val fromIndex = states.indexOfFirst { it.name == name }
+            if (fromIndex == -1) return@update states
             val mutableStates = states.toMutableList()
             val item = mutableStates.removeAt(fromIndex)
-            val adjustedToIndex = if (toIndex > fromIndex) toIndex - 1 else toIndex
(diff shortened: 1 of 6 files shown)


sproctor:
Both review comments addressed in 69028fdb.

**`pointerInput` keyed on the index (major)** — fixed, and it turned out to matter more than a restart nuisance. Each window in the dock is wrapped in `key(uiState.name)`, so its composition identity survives a neighbor being removed; the index in the `pointerInput` key was therefore the *only* thing killing an in-flight drag when a window ahead of the dragged one left the dock. That is the same mid-drag dock change this PR exists to handle, so it was cancelling exactly the drags that most needed to complete. Both handlers now key on `uiState.name` and `location`, with a comment on why the index must stay out.

This is also the flip side of the crash: with the index keyed, a window removed *before* the dragged one silently cancelled the gesture, while one removed *after* left the index untouched, let the gesture finish, and fed stale bounds into `removeAt`/`add` — the `Index: 2, Size: 1` in the report.

Since that key was the last use of the parameter, `WindowViewSlot` no longer takes an `index` at all, in the same spirit as dropping `DropResult.sourceIndex`.

**Skip persistence when the savable order is empty (minor)** — fixed; `changeWindowPositions` now returns before the transaction when the dock holds nothing savable. Worth noting the DAO call was not a pure no-op in that case: with an empty `names`, `setPositions` still walked the dock and compacted any sparse positions, so a reorder of transient-only panels could rewrite saved rows (including closed windows' remembered slots) that the user never touched. Returning early avoids the round trip and that side effect.

`./gradlew jvmTest` passes and compose compiles for Android.

Issue #4, opened 2022-07-10
Aliases


sproctor:
replace text from entry on submission

sproctor:
done

PR #283, opened 2026-08-17
Open a window the game detaches in a window of its own

sproctor:
`location='detach'` now opens the window detached, instead of docking it into the band above main.

## Why it was docked

The mapping from protocol location to dock area carried this comment:

> We dock rather than float, so the locations the real client floats ... share the band above main with the room window.

That was written when nothing could float. Windows can be detached now (#279), so the announcement can be honoured — and the real client does honour it: its saved settings carry such a window as `frame='float'` with a size of its own (`<w id='scharsheet' frame='float' location='detach' detach='t'>`).

## Scope

- **Default only.** `reconcile` places nothing that's already in the layout, so a character who has docked the window keeps it docked across a reconnect. Detaching is what the game asks for on a window nobody has arranged yet, not something re-imposed every connect.
- **Never main** — it's the fixed point the areas are measured against.
- **Never where the platform has no OS windows.** Mobile falls back to the band above main exactly as before, rather than dropping the announcement.

The location's *anchor* still maps to the centre band, and deliberately so: that's where a detached window belongs whenever it's docked rather than floating — on reattaching, or on a platform without windows. Where it opens and which area it belongs to once docked are separate questions, which the comment there now says.

## Testing

- `aWindowAnnouncedDetachedOpensInItsOwnWindow` — confirmed to fail on the old behaviour and pass on the new.
- `aDetachedWindowTheUserHasDockedStaysDocked` — guards the "arrangement wins over announcement" property through a save/restore round trip. It passes either way; it's there so a future change can't quietly start re-detaching a window the user docked.

Verified in the running app against a replayed session: a `<streamWindow location='detach'>` came up as a separate OS window titled "Character Sheet", carrying its stream content and dock header, while a `location='left'` window in the same session still docked into the left column and the detached one stayed out of the main window entirely.

## Summary by CodeRabbit

* **New Features**
  * Windows marked as detached now open in separate desktop windows when supported.
  * Regular windows continue to open in their designated docked areas.
  * Detached windows are excluded from the main window and remain docked on unsupported platforms.
* **Bug Fixes**
  * Detached windows that users manually redock remain docked after saving, restoring, or reconnecting.
* **Documentation**
  * Clarified how detached windows are positioned when docked.


sproctor:
Updated to compose-docking **0.3.0** in 01618dfb.

That release fixes the thing this change makes everyone notice: a detached window couldn't be dragged back. Starting the drag undocked the window, which emptied the detached window and had it collected mid-gesture, taking the gesture with it — sproctor/compose-docking#4 carries the window itself instead, so the panel rides along inside the real window and dropping it over a dock area docks it there. 0.2.2 is in here too (the stray divider grip down the window edge).

**One behaviour change worth knowing:** detached windows are undecorated as of 0.3.0, and the window's own header is the whole of its chrome — one title bar instead of two, in the renderer's colours rather than the window manager's. Dragging that header moves the window. Nothing needed changing on our side for it: the header already carried the `...` menu and the close button, since the library draws no header buttons of its own and never did.

There is no minimize on a detached window now. The library deliberately draws none, leaving it to the app the way a dockable's `trailingActions` already are; worth adding later if it's wanted, not a blocker here.

## Verified

`./gradlew check -PiosSkip=true -PlintSkip=true` passes, and in the running app against a replayed session:

- a `<streamWindow location='detach'>` opens as its own window, undecorated, carrying its stream and its `...`/`×`
- dragging that header onto the main window's dock area docks it back — it lands as a pane below `main` and the separate window is gone
- a `location='left'` window in the same session still docks as before

One of my earlier drag attempts appeared to fail; that was my drop point going stale after you moved the main window, not the library.

PR #284, opened 2026-08-17
Let a skin size a progress bar, over the panel font setting

sproctor:
A skin could say what colour a progress bar is but not what size its label is, so every bar took the panel font — a preference about panel text in general — and a skin that had laid out a bar to fit a particular slot had no say in whether the text would fit it.

## Precedence

Skin fonts now sit between the panel font setting and the user's own font for the bar:

| | |
|---|---|
| user's font for this bar | wins over everything — they can see what they're changing |
| **the skin** | new; wins over the panel font, because a skin that sizes a bar knows the slot it laid out |
| panel font setting | a general preference about panel text |
| platform compact base | |

Resolved **field by field**, matching how `withFont` already merges, so a skin setting only a size keeps the family and weight it would otherwise have had. And taken from the bar's own skin entry before the shared `progressBar` entry — a skin can size all its bars in one place and still say something different about one of them.

The `fontFamily`/`fontSize`/`fontWeight` fields have been on `SkinObject` all along, accepted and ignored.

## The minivitals

Given 11sp in the bundled skin, which is what wanted saying in the first place. They sit in a fixed-height row drawn as chrome rather than as a window, so they never took the panel font at all and were stuck at the 9sp base.

The skin change is five lines — `fontSize` on `HealthBar > health` and its four siblings, which is how the skin already addresses those bars (a bar's entry is a child of `<id>Bar` keyed by its id). `HealthBar2 > health2` is deliberately untouched.

## Testing

`ProgressBarSkinFontTest` covers the resolver: nothing set contributes nothing, the shared entry applies to a bar that says nothing itself, the bar's own entry wins over the shared one, a bar setting only a size keeps the rest, and a bar's entry alone works with no shared entry. The resolver is a pure function over the skin map with a thin composable wrapper, so this is testable without a composition.

Verified in the running app with a `location='statBar'` minivitals panel. A/B against the skin entry removed shows the labels visibly larger with it, confirming the skin value — not the base — is what's driving the size.

## Summary by CodeRabbit

* **New Features**
  * Added skin-based font styling for health, mana, stamina, spirit, and concentration bars.
  * Progress-bar text now supports shared defaults, individual bar overrides, and field-level customization.
  * Set the default progress-bar text size to 11 for improved readability.
  * Applied consistent font resolution across desktop and mobile interfaces.

* **Bug Fixes**
  * Improved progress-bar text styling by correctly combining panel, skin, and saved customization settings.

### DIFF
--- a/compose/skin/skin.json
+++ b/compose/skin/skin.json
@@ -17,6 +17,7 @@
   "HealthBar": {
     "children": {
       "health": {
+        "fontSize": 11,
         "top": 0,
         "left": 0,
         "color": "#FFFFFF",
@@ -39,6 +40,7 @@
   "ManaBar": {
     "children": {
       "mana": {
+        "fontSize": 11,
         "top": 0,
         "left": 0,
         "color": "#FFFFFF",
@@ -50,6 +52,7 @@
   "StaminaBar": {
     "children": {
       "stamina": {
+        "fontSize": 11,
         "top": 0,
         "left": 0,
         "color": "#FFFFFF",
@@ -61,6 +64,7 @@
   "SpiritBar": {
     "children": {
       "spirit": {
+        "fontSize": 11,
         "top": 0,
         "left": 0,
         "color": "#FFFFFF",
@@ -72,6 +76,7 @@
   "ConcentrationBar": {
     "children": {
       "concentration": {
+        "fontSize": 11,
         "color": "#FFFFFF",
         "bar": { "light": "#109A88", "dark": "#22C7AE" },
         "background": { "light": "#C6C6BE", "dark": "#2C2F34" }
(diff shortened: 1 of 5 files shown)


sproctor:
Addressed in 1a7c80aa — the one nitpick was valid and worth having.

Every test used a lowercase `progressBar` key, so nothing pinned the `getIgnoringCase` lookup: switching it to an exact `skin["progressBar"]` would have gone unnoticed. Added `theSharedEntryIsFoundWhateverCaseItIsWrittenIn` with a `"PrOgReSsBaR"` key.

Confirmed the test actually has teeth by temporarily swapping the resolver to an exact lookup — it fails there and passes with `getIgnoringCase`, so it pins the contract rather than just restating a passing case.

Worth noting *why* this isn't incidental, which came out of a separate investigation with the Wrayth bench: skins and the server disagree about case as a matter of course. GS4 sends `name='healthBar'` while `storm.skn` defines `HealthBar`, and the real client skins those bars regardless — verified as a genuine case-insensitive compare by sending `HeAlThBaR` and `sTaMiNaBaR` and still getting the health bar's red and the stamina bar's gold. So case-insensitive lookups here match the client we're following, and the new test guards that.

sproctor:
@coderabbitai review

PR #289, opened 2026-08-17
Hold the line's text and its open component as one value

sproctor:
Answers the `// TODO: separate buffer into its own class` above `flushBuffer`. Yes — but the reason turned out to be stronger than tidiness: the invariant a class would enforce is already broken, and breaking it takes the main window out for the rest of the session.

## What was wrong

Three fields had to agree and nothing made them:

- `bufferText` routed on `elementBuffer != null`
- `flushBuffer` gated on `componentId == null`
- two handlers cleared **`componentId` alone** — `WraythStreamEvent` and `WraythPromptEvent`

Once those two disagree they never re-converge. Every later line accumulates into an element buffer nothing flushes, `streamBuffer` stays null, and `flushBuffer` cheerfully flushes nothing.

Driven through the client on the bench:

```
BEFORE visible
<component id='x'>foo<prompt time="...">&gt;</prompt>bar</component>
AFTER-A does this appear?
AFTER-B does this appear?
AFTER-C does this appear?
```

`BEFORE` renders. **A, B and C never appear, and nothing appears ever again.** After the fix all three render, and `bar` lands on the line instead of vanishing.

Two honest limits on the severity. It needs a reset event *inside* a component on one line, which is malformed — no real server sends that. And an unterminated `<component>` spanning a line does **not** trigger it, because `handleContent` rule #5 closes open tags at end of line; I checked that case first and it was clean. What makes it worth fixing anyway is that `WraythProtocolHandler` carries five numbered rules for malformed tags precisely because malformed input turns up, and going permanently blind is a poor way to meet it.

## The change

`LineBuffer` holds the line's text and the open component as one value, so the id and its content cannot come apart. The operations are the ones that actually exist — `append`, `openComponent`, `closeComponent`, `discardComponent`, `takeLine` — and both `closeComponent` and `discardComponent` end with the line gathering its own text again, which is what the old pair of assignments was reaching for and only half did.

Two things it deliberately does not do:

- **It doesn't own the flush.** `flushBuffer` needs `currentStream`, the windows map, `isPrompting` and logging; the class returns the text and lets the client place it, or half the client comes with it.
- **`takeLine()` returns an empty `StyledString`, not null, for an empty line** — a blank line is still a line. Null is reserved for "a component is open, don't flush yet", which is the old `componentId == null` guard.

## Testing

10 tests over the contract, including the discard path that the prompt and stream handlers take. Plus the bench sequence above, and a check that ordinary rendering is unchanged — a `<compDef>` placeholder still fills from a later `<component>`.

**Performance:** A/B'd against the same tree without this commit using `:compose:streamNetworkBenchmark` (2 connections, 20k lines, 7 interleaved rounds each). No measurable difference — median 853.6ms vs 830.4ms, but run-to-run stdev is ~110ms and the paired per-round deltas swing ±245ms with a mean of +4ms, so the harness cannot resolve a difference this small. That matches what the code says it should be: the line path allocates exactly as before, and only appends *inside* a component allocate one extra short-lived object, on a path that already allocates a `StyledString` per append.

Note this benchmark only became meaningful again with #288 — before that it ran the client in raw mode start to finish.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of streamed text and styled components.
  * Prevented incomplete component content from being displayed as a finished line.
  * Preserved blank lines and ensured discarded components no longer appear.
  * Improved recovery when components are closed or interrupted.

### DIFF
--- a/wrayth/src/commonMain/kotlin/warlockfe/warlock3/wrayth/network/LineBuffer.kt
+++ b/wrayth/src/commonMain/kotlin/warlockfe/warlock3/wrayth/network/LineBuffer.kt
@@ -1,0 +1,68 @@
+package warlockfe.warlock3.wrayth.network
+
+import warlockfe.warlock3.core.text.StyledString
+
+/**
+ * The text being assembled for the line currently being parsed, and the component being assembled
+ * inside it, if any.
+ *
+ * A component's id and its content are one value here so they cannot get out of step. Two questions
+ * are asked of this state by different callers - where appended text goes, and whether the line may
+ * be flushed yet - and both answers turn on whether a component is open. Held apart, a caller could
+ * answer them differently, and text would accumulate somewhere nothing ever flushed.
+ */
+internal class LineBuffer {
+    private var line: StyledString? = null
+    private var component: OpenComponent? = null
+
+    private data class OpenComponent(
+        val id: String,
+        val text: StyledString,
+    )
+
+    /** Appends to the open component if there is one, and to the line otherwise. */
+    fun append(text: StyledString) {
+        val open = component
+        if (open != null) {
+            component = open.copy(text = open.text + text)
+        } else {
+            line = line?.plus(text) ?: text
+        }
+    }
+
+    /** Starts a component, so that [append] gathers its content instead of the line's. */
+    fun openComponent(id: String) {
+        component = OpenComponent(id, StyledString())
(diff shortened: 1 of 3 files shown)


sproctor:
@coderabbitai review

sproctor:
@coderabbitai review

PR #291, opened 2026-08-17
Show an unhandled tag once, when debug output is on

sproctor:
Two small things: `pulse` joins the ignored tags, and an unknown tag now says so on screen once, when debug output is on.

## Unknown tags were invisible from inside the game

`warnOnUnknownTag` already logs each new one and the client handler was a commented-out `// debug(...)`. The log is not where anyone is looking when something the game sent failed to appear — and from inside the game there was nothing to see at all.

With debug output on (`-d`, or any dev build, which is how `isDev` already sets Kermit's min severity) the **first sighting of each tag name** now prints to the main stream in the error style:

```
before the unknown tags
unhandled tag: <neverSeenBefore>     <- red, once
text after it
second time, should be silent
third time, should be silent
pulse should be silent
unhandled tag: <alsoUnknown>         <- red, its own first sighting
a different tag, should report once
```

## The flag comes from the parser, not a second tally

`WraythUnhandledTagEvent` now carries `firstSighting`, set by `warnOnUnknownTag`, which was already returning early for names it had seen. It also already goes quiet past `MAX_UNKNOWN_TAGS_REPORTED` (256) distinct names, so a server generating tag names per line cannot flood the stream — that cap now covers the on-screen message for free.

Keeping a separate set in `WraythClient` would have been a second source of truth for the same question, with its own chance to disagree about the cap.

The event is still emitted every time, so nothing downstream has to guess whether a tag it ignored was seen; only the flag is once.

## Testing

6 tests: first sighting, repeats not, each distinct name getting its own, `pulse` ignored rather than unhandled, and the cap suppressing the flag beyond 256 names.

Verified in the app against the bench with `-d`, which is the screenshot transcribed above.

**One gap, stated rather than papered over:** the debug-*off* path is not reachable from a dev build, because `isDev` forces debug on, and forcing a non-dev version to test it would trip Sentry initialisation. So what is covered instead is the direction of the `minSeverity <= Severity.Debug` comparison — the part that could invert silently and show these only in the builds meant not to.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of previously unknown protocol tags.
  * Prevented repeated reports for the same tag.
  * Ignored expected `pulse` messages.
  * Limited unknown-tag reporting to avoid excessive announcements.
  * Added clearer debug-level error messages for newly encountered tags.

* **Tests**
  * Added coverage for first-time, repeated, and distinct unknown tags.
  * Verified reporting limits, ignored messages, and debug severity behavior.

### DIFF
--- a/wrayth/src/commonMain/kotlin/warlockfe/warlock3/wrayth/network/WraythClient.kt
+++ b/wrayth/src/commonMain/kotlin/warlockfe/warlock3/wrayth/network/WraythClient.kt
@@ -1,6 +1,7 @@
 package warlockfe.warlock3.wrayth.network
 
 import co.touchlab.kermit.Logger
+import co.touchlab.kermit.Severity
 import com.eygraber.uri.Uri
 import kotlinx.collections.immutable.mutate
 import kotlinx.collections.immutable.persistentListOf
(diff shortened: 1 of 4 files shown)


sproctor:
@coderabbitai review

PR #292, opened 2026-08-18
An empty string literal is an empty string, not a crash

sproctor:
Sentry `DESKTOP-3N` / issue 7674173141, against 3.0.166. **It still applies to main**, reproduced rather than assumed.

## Reading the report

The stack points at `WslScript.kt` line **701**, and the file at `v3.0.166` is **694 lines long**. That is `reduce` being an inline function: the `throw` inside it belongs to kotlin-stdlib but is attributed to the calling file, at a synthetic line past its end. So the crashing frame is the function *containing* the reduce, and the frame under it identifies which:

| stack frame | `v3.0.166` source |
|---|---|
| `getValue` @ 701 | inlined `reduce`, past EOF |
| `getValue` @ 595 | `WithValue.getValue` → `valueExpression.getValue(context)` |

The only `WslValueExpression` with a reduce is `WslStringExpression` (line 615 there, 651 on main). Reproducing on main gives the identical pair:

```
UnsupportedOperationException: Empty collection can't be reduced.
  at WslValueExpression$WslStringExpression.getValue(WslScript.kt:739)   <- past EOF again
  at WslPrimaryExpression$WithValue.getValue(WslScript.kt:631)
```

## The bug

```
stringLiteral : QUOTE_OPEN stringContent* QUOTE_CLOSE;
```

`stringContent*` is **zero or more**, so `""` parses to an empty content list, and `reduce` on it throws. Any script evaluating an empty string literal — `if "" then ...` is enough — kills the script with an uncaught `UnsupportedOperationException`, which is why this arrived as a fatal rather than a script error.

`WslCommand.execute` already uses `reduceOrNull` a few hundred lines above, for exactly this reason; this brings the string expression in line with it.

## Scope

The two neighbouring reduces are safe and left alone. `disjunction: conjunction (OR conjunction)*` and `conjunction: equality (AND equality)*` both take **one** or more, so neither list can be empty.

## Testing

Three tests, all confirmed to fail without the fix (`Empty collection can't be reduced.`):

- `anEmptyStringLiteralIsAnEmptyString` — `if "" then echo hi` evaluates to `""`
- `anEmptyStringConcatenatesWithText` — `"" + "abc"` is `"abc"`
- `aNonEmptyStringStillWorks` — the ordinary path is unchanged

Two assertions I tried first were wrong about the language rather than the bug, and are not in the final tests: `%{...}` is not valid inside a string literal, and an empty string is not coercible to boolean (`WslRuntimeException: String that is not "true" or "false" cannot be used as a boolean`), so `if ""` still fails — as a script error, not a process kill.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed empty string literals so they evaluate correctly instead of causing errors.
  - Empty strings can now be safely concatenated with other text.
  - Existing non-empty string behavior remains unchanged.

- **Tests**
  - Added coverage for parsing, evaluating, and concatenating empty string literals.

### DIFF
--- a/scripting/src/commonMain/kotlin/warlockfe/warlock3/scripting/wsl/WslScript.kt
+++ b/scripting/src/commonMain/kotlin/warlockfe/warlock3/scripting/wsl/WslScript.kt
@@ -645,10 +645,12 @@ sealed class WslValueExpression {
         val content: List<WslStringContent>,
     ) : WslValueExpression() {
         override fun getValue(context: WslContext): WslValue {
+            // `stringContent*` in the grammar takes zero or more, so "" has nothing to reduce.
             val value =
                 content
                     .map { it.getValue(context) }
-                    .reduce { acc, s -> acc + s }
+                    .reduceOrNull { acc, s -> acc + s }
+                    ?: ""
             return WslString(value)
         }
     }
--- a/scripting/src/commonTest/kotlin/wsl/WslEmptyStringTest.kt
+++ b/scripting/src/commonTest/kotlin/wsl/WslEmptyStringTest.kt
@@ -1,0 +1,64 @@
+package warlockfe.warlock3.scripting.wsl
+
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import kotlinx.io.buffered
+import kotlinx.io.files.Path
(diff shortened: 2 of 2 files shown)


sproctor:
@coderabbitai review

PR #293, opened 2026-08-18
Get the iOS app building, reporting crashes, and ready to submit

sproctor:
Gets the iOS app to the point where it can be archived and uploaded to App Store Connect. Six commits, each independently reviewable.

### The iOS build was broken

CI runs `-PiosSkip=true` because the Linux runner can't compile Apple targets, so nothing exercised iOS and it rotted three separate ways:

- `WarlockSettingsSync` used `toSortedSet()`/`toSortedMap()`, both JVM-only. Every one of the dozen `'Char' vs 'String'` errors was a cascade off these two.
- Three files used `Dispatchers.IO` without importing it. It *is* available in `commonMain` — coroutines declares it in `concurrentMain` as an extension property, so it needs `import kotlinx.coroutines.IO`.
- The Xcode build phase didn't pass `-PiosSkip=false`, and it skipped the Compose resource sync whenever the IDE set `OVERRIDE_KOTLIN_BUILD_IDE_SUPPORTED` — so running from IntelliJ produced an app with an empty `compose-resources` that died on the first `Res.readBytes`.

kompress 2.3.1 also moved `Inflater` and replaced `inflate(data, raw = true)` with `decompress(data)`. `ZipReader` is the skin loader, so that one would have broken skins at runtime, not just compilation.

`iosSkip` now defaults to false, so `./gradlew check` covers iOS locally.

### Crash reporting on Android and iOS

Desktop already reported to Sentry; the mobile apps didn't. All three now share one project with per-platform releases (`android@3.1.0`).

iOS deliberately does **not** use the Kotlin SDK. Its cinterop carries a hardcoded `-framework Sentry` that Xcode satisfies via SPM but Gradle can't when linking the iOS test binary — it broke `./gradlew check` outright. iOS starts the Cocoa SDK from `iOSApp.swift` instead; the crash handler is process-wide, so Kotlin/Native crashes are still caught. Cost: the DSN appears in two files, each pointing at the other.

### A use-after-free in the iOS TLS path

Both `CFArrayCreate` calls in `evaluateServerTrust` passed `null` for callbacks, so the arrays never retained the cert and policy — which were then released while `trust` still referenced them. This is consistent with a handshake that failed `errSecParam` once and worked on retry.

### App Store prep

The app icon was still the blue Compose template cube. Rebuilt from the 1024px hat in `icons/icon.icns` on the same background the Android icon uses, flattened because Apple rejects icons with an alpha channel. Plus `PrivacyInfo.xcprivacy` and an export-compliance declaration.

### CI

New macOS job: iOS tests against Kotlin/Native, then an unsigned `xcodebuild`. The Gradle task alone wouldn't have caught any of the three failures above — the Xcode step covers the build phase, resource sync, SPM resolution, and linking.

### Verified

`./gradlew check -PlintSkip=true` green with iOS enabled, `iosSimulatorArm64Test` green, Xcode Release build green, and a build has been successfully uploaded to App Store Connect.

## Summary by CodeRabbit

* **New Features**
  * Added production crash reporting for Android, desktop, and iOS with privacy-conscious settings.
  * Added iOS privacy and export-compliance declarations.
  * Added automated iOS simulator testing and app builds.

* **Bug Fixes**
  * Improved iOS archive extraction for compressed files.
  * Improved secure connection handling and resource cleanup.
  * Improved synchronization ordering consistency.

* **Improvements**
  * Enabled iOS builds by default.
  * Updated iOS branding to “Warlock FE.”
  * Updated compression support for improved compatibility.


sproctor:
Addressed the two outside-diff findings.

**Fixed — TLS/socket resource release** (20df1703)

All four parts were valid, and there was a fifth leak nearby:

- `createAndConnectSocket` left the descriptor open when `connect()` failed. Not flagged, but the worst of the set: the client retries, so every failed attempt burned an fd.
- `setupTLS`/`setupDefaultTLS` threw out of a failed handshake without releasing the SecureTransport context or disposing the `StableRef` the IO callbacks read the fd from. A failed handshake is routine here, not exceptional.
- Those throws also stranded the connected socket in the callers, which only close it once a `TLSSocketConnection` exists.
- The close callbacks called `SSLClose` but never `CFRelease`. `SSLCreateContext` returns a +1 reference and `SSLClose` does not consume it, so every cleanly closed connection leaked a context.

**Not changing — `Apple Distribution` in the Release configuration**

The reasoning is sound from the diff alone, but a build from this branch has already been archived and uploaded to App Store Connect successfully with `CODE_SIGN_IDENTITY = "Apple Development"` in place. `CODE_SIGN_STYLE = Automatic` means Xcode resolves the distribution identity at export time, which is what happened.

Pinning `Apple Distribution` would also break running a Release build on a local device, which needs a development identity. So it would fix nothing observable and cost something real. Worth revisiting only if an export actually fails on the identity.

sproctor:
Fixed in 76456594 — and this one was worse than it looked, partly because of my previous commit.

`scope.cancel()` does not wait, and both IO jobs sit inside blocking native calls that never observe cancellation until they return. So a job could still be inside `SSLRead` while `SSLClose`, `CFRelease(sslCtx)` and `stableRef.dispose()` ran underneath it. Adding the `CFRelease` in 20df1703 turned this from "reads a closed context" into a genuine use-after-free.

Teardown is now ordered:

- `shutdown(fd, SHUT_RDWR)` instead of `close(fd)`, to unblock the read/write calls the jobs are parked in. `shutdown` leaves the descriptor number allocated — closing it here would let another socket take the number while the IO callbacks still hold it and write to the wrong connection.
- cancel the job, then release the TLS context and close the descriptor from `invokeOnCompletion`, which runs only after both children have finished.

Applied to `openPlainSocket` as well. It has no context to free, but the same fd-reuse hazard.

sproctor:
Fixed in the latest commit.

Correct as described: all three close closures used `var closed` with a plain check-then-set, and `close()` is reachable from more than one thread. Two callers could both pass the check. For the plain socket that double-closes a descriptor; for the TLS sockets it double-releases the SecureTransport context and disposes the stable ref twice — both crashes rather than leaks, so this is worse than a redundant cleanup.

Now uses `kotlin.concurrent.atomics.AtomicBoolean.compareAndSet(false, true)`, which hands the gate to exactly one caller.

sproctor:
@coderabbitai resume

sproctor:
@coderabbitai review

PR #294, opened 2026-08-19
Rewrite the iOS sockets on Network.framework

sproctor:
Replaces the hand-rolled SecureTransport implementation of the iOS sockets with Network.framework.

### Why

Apple deprecated SecureTransport years ago, and every review finding on #293 was a lifetime or concurrency defect in the old design — CoreFoundation objects released while still referenced, a socket and TLS context stranded on failure paths, teardown racing two jobs parked inside blocking native calls, a non-atomic close gate. Each fix exposed the next layer.

Network.framework removes the category rather than patching it. `nw_connection_t` and friends are Objective-C objects, so the Kotlin/Native runtime owns their lifetime and there is nothing to `CFRelease`. Cancellation is asynchronous by design. Both IO paths are callbacks bridged into coroutines, so no thread parks in a blocking call that cancellation cannot interrupt.

Certificate pinning keeps the previous semantics exactly: pinned certificate as an additional anchor, system anchors still trusted, hostname checking off because the Simutronics CA certificate has no SAN for the game host.

### Three interop traps

None of these produce a compiler error. All three compiled cleanly and failed at runtime.

- **`NW_PARAMETERS_DISABLE_PROTOCOL` is matched by pointer identity.** Kotlin/Native wraps a block crossing the boundary, so the comparison never matches and the framework *calls* the sentinel as an ordinary configuration block — which had a **plain** socket attempting TLS. The protocol stack is assembled by hand instead, which also means SNI must be set explicitly.
- **`NW_CONNECTION_DEFAULT_MESSAGE_CONTEXT` is also a sentinel**, not a real object. Converting it aborts the process inside `Kotlin_Interop_refFromObjC`.
- **`dispatch_data_create_map` returns an object owning the flattened buffer.** Left unused, the runtime can collect it mid-`memcpy`.

### Also fixed

The IO scope used a parentless `SupervisorJob`, and `coroutineContext + job` *replaces* whatever `Job` the context held — so the loops were detached from the caller's cancellation and the connection stayed open until something called `close()` explicitly. Carried over from the old implementation, where it was worse: the stranded loops held a file descriptor, an SSL context and a stable ref. The job is now a child of the caller's, and teardown runs from the job's completion as well as from `close()`.

### Verification

- **Plain TCP**: verified end-to-end against a real server — connect, send, and a genuine HTTP response read back.
- **Pinned TLS**: verified by connecting to a game. That path is `openTLSSocket`, since iOS `SgeSettings` sets `secure = true`.
- **`openDefaultTlsSocket`**: not exercised. It is used only by the MUD Mobile router and differs from the pinned path only in adding no verify block.

Worth knowing for future debugging: a Kotlin/Native test binary cannot do TLS trust evaluation or reach the Keychain — it has no app container, so Security services are restricted. TLS there fails with `-9808 "bad certificate format"` regardless of the code. Verification of anything touching Security has to happen in the app.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iOS connection reliability for plain, standard TLS, and certificate-pinned connections.
  * Improved handling of connection failures, cancellations, interrupted network activity, and closed connections.
  * Strengthened certificate validation, including support for system trust anchors and PEM certificates.
  * Improved asynchronous sending and receiving, with more consistent reporting of network errors.
  * Fixed handling of data received alongside connection closures, helping prevent lost messages.

### DIFF
--- a/wrayth/src/iosMain/kotlin/warlockfe/warlock3/wrayth/util/SocketHelpers.ios.kt
+++ b/wrayth/src/iosMain/kotlin/warlockfe/warlock3/wrayth/util/SocketHelpers.ios.kt
@@ -1,281 +1,416 @@
-@file:Suppress("DEPRECATION")
-@file:OptIn(kotlinx.cinterop.ExperimentalForeignApi::class, ExperimentalAtomicApi::class)
+@file:OptIn(ExperimentalForeignApi::class, ExperimentalAtomicApi::class)
 
 package warlockfe.warlock3.wrayth.util
 
+import co.touchlab.kermit.Logger
 import io.ktor.network.selector.SelectorManager
 import io.ktor.utils.io.ByteChannel
 import io.ktor.utils.io.ByteReadChannel
 import io.ktor.utils.io.readAvailable
 import io.ktor.utils.io.writeFully
 import kotlinx.cinterop.COpaquePointer
-import kotlinx.cinterop.CPointer
-import kotlinx.cinterop.CPointerVar
-import kotlinx.cinterop.StableRef
-import kotlinx.cinterop.ULongVar
+import kotlinx.cinterop.COpaquePointerVar
+import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.cinterop.addressOf
 import kotlinx.cinterop.alloc
 import kotlinx.cinterop.allocArrayOf
-import kotlinx.cinterop.asStableRef
 import kotlinx.cinterop.convert
 import kotlinx.cinterop.memScoped
-import kotlinx.cinterop.nativeHeap
-import kotlinx.cinterop.pointed
 import kotlinx.cinterop.ptr
 import kotlinx.cinterop.reinterpret
-import kotlinx.cinterop.staticCFunction
 import kotlinx.cinterop.usePinned
 import kotlinx.cinterop.value
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.IO
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancel
 import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.suspendCancellableCoroutine
 import platform.CoreFoundation.CFArrayCreate
 import platform.CoreFoundation.CFArrayRef
 import platform.CoreFoundation.CFDataCreate
 import platform.CoreFoundation.CFErrorRefVar
 import platform.CoreFoundation.CFRelease
 import platform.CoreFoundation.kCFTypeArrayCallBacks
-import platform.Security.SSLClose
-import platform.Security.SSLConnectionRef
-import platform.Security.SSLConnectionType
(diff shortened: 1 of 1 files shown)


sproctor:
All four addressed in the latest commit. Two of them were pinning failing open, which is the worst shape this file could fail in.

**TLS option handles (critical).** `nw_tls_create_options` and `nw_tls_copy_sec_protocol_options` were allowed to return null with the connection carrying on regardless — no server name, and on the pinned path no verify block. A connection that looks fine while doing none of the checking it was asked for. Both are `checkNotNull` now, at the stack-building site and in the pinned path's configure block, so setup fails rather than proceeding unverified.

**SecTrust OSStatus.** Correct, and worth spelling out why it matters: if a setter fails, `SecTrustEvaluateWithError` evaluates against the default configuration and can return `true` for a certificate that does not chain to the pinned one. Any non-success now denies trust and logs.

**Write loop.** Cancelled with the cause. As noted, `SgeClientImpl` writes from its protocol loop, so this turned a send failure into a silent hang rather than an error.

**Receive error.** Now propagates rather than being mapped to `null`, and the read channel is cancelled with the cause, so a dropped connection is no longer indistinguishable from a clean logout.

One detail the suggestions did not cover: both loops have to tell an intentional close from a real failure, because cancelling an `nw_connection_t` makes outstanding receives complete with an error too. Reporting that as a failure would make every normal disconnect look like a network error. Both loops now consult the same atomic close gate that guards teardown, and only cancel the channel with a cause when the failure did not come from us.

`./gradlew check -PlintSkip=true` and the Release Xcode build are green.

sproctor:
Fixed in 46da0e02.

`toByteArray` returned an empty array when `dispatch_data_create_map` failed, which is indistinguishable from a legitimately empty payload — the read loop would treat lost bytes as an ordinary empty read and carry on. It returns null now, and `receiveChunk` turns that into an `IOException`, so data that arrived but could not be read out surfaces as a read failure instead of vanishing.

Closing out the two open questions from the PR description, both decided deliberately:

- **Trust configuration stays as it is.** `SecTrustSetAnchorCertificatesOnly(trustRef, false)` and the basic X.509 policy preserve the behaviour of the SecureTransport implementation this replaces. Changing it is a separate decision about the app's trust model, not part of porting the transport.
- **`openDefaultTlsSocket` remains unexercised.** It is only used by the MUD Mobile router, which is out of scope here.

sproctor:
Half of this is fixed in 68cb8389; the other half is not implementable in this ktor version.

**Content delivered with an error — fixed.** The `error != null` branch was checked before the content, so a receive that delivered both the last bytes and a failure dropped the bytes. The content is now mapped first and carried on `Received` with the failure attached, and the read loop writes it before raising.

**`readChannel.close(receiveError)` — no such call.** ktor 3.5.2's `ByteChannel` offers `close()`, `flushAndClose()` and `cancel(cause)`, and only the last carries a reason; there is no close-with-cause that would preserve buffered bytes. Checked against the `ktor-io` klib rather than assuming.

The closest available behaviour is what the fix does: write the content and rely on the channel autoflushing so it reaches the reader, then cancel with the cause. Bytes still buffered at that moment are discarded, which is the residual gap. Trading `cancel(cause)` for `flushAndClose()` would preserve them but throw away the error signal entirely — the opposite failure, and worse for a protocol where a dropped connection has to be distinguishable from a clean logout.

PR #301, opened 2026-08-21
Give the phone its own tab bar, and share the window view between platforms

sproctor:
## Why

The phone game screen showed whatever windows the game had announced, in arrival order, with no way to add, remove or reorder them. A window the game closed was unreachable — the only route to one was Settings → Windows → Show/Hide.

## The phone tab bar

- A **wide navigation rail** (`ModalWideNavigationRail`, `hideOnCollapse = true`) lists the character's windows and adds them as tabs. Windows already tabbed show as selected, so tapping one is a shortcut to its tab.
- **Tabs reorder by long-press drag.** Long press has to be the drag gesture: press-and-move stays free for scrolling the strip when tabs overflow.
- **Dragging a tab onto a floating target removes it.** The reorder library has no external drop target, and a horizontal drag only translates the tab along x — so the tab never travels toward the target and the hit test needs the raw pointer, tracked on `PointerEventPass.Initial` from the strip root. No reorder suppression is needed: the drag has already reordered, and removing an item leaves everything else in the order it reached.
- **The selected tab carries a `⋮`** with Window settings / Clear window / Close, mirroring what `MobileDockWindowActions` already offers per docked window.
- **Order is saved per character** under a new `phoneTabOrder` key.

The phone keeps this list separate from the `WindowSettings.open` flag that feeds the docking layout — adding a tab on a phone must not rearrange a dock that size never shows. A character with nothing saved keeps every window as a tab, exactly as before; the first edit makes the list explicit. Chrome panels and transients are now filtered out; they could previously appear as tabs that led nowhere.

16 unit tests cover the model in `PhoneTabOrderTest`.

**Tablet layout dropped** — Expanded widths use the docking layout. That retires `PaneLocation`, `setLocationSize`, the tablet pane-location setting and the four pane width/height flows.

## Sharing what was duplicated

| Was | Now |
|---|---|
| `DesktopWindowView` + `WindowView`, identical signatures kept in step by hand | `WindowView` expect/actual |
| Two byte-identical `DockedWindow` copies | one in `commonMain` |
| `DesktopTabActivityDot` | shared `TabActivityDot` — mobile gains the unread-text dot it was missing |
| `GameWindowStyles` in jvmMain, its 8 locals re-inlined on mobile | one copy in `commonMain` |
| The same 22-line `onPreviewKeyEvent` handler in both game views | `rememberGameKeyHandler` |
| `NavMenuItem` and `ActionNavMenuItem`, 57 identical lines | `WarlockNavMenuItem` |

## Composition locals: eight to five

`LocalBaseStyle` now carries the whole `ResolvedStyle` rather than a lossy projection, so `LocalDefaultFont` — which existed only to carry back the font half `StyleDefinition` cannot hold — is gone. `LocalMonoFont` is gone too; nothing ever read it (the mono font reaches the renderer through `WindowRegistryImpl.monoFontFor`). `LocalPanelFont` + `LocalPanelScale` become `LocalPanelDefaults`.

## Android system bars

`AppTheme` is `android:Theme.Material.Light.NoActionBar`, which leaves `windowLightStatusBar` false — so the status-bar icons were white. Harmless while that bar had a background of its own; invisible once targetSdk 37 forced edge-to-edge over light content. `MainActivity` now calls `enableEdgeToEdge()` and drives the icon appearance from the app's own resolved theme, which the automatic handling gets wrong whenever `ThemeSetting` disagrees with the system.

## Shim cleanup

`WarlockResizablePanel` (168 lines), `WarlockGroupHeader`, and the bare `WarlockCheckbox`/`WarlockRadioButton` had **no call sites at all**. The two genuine pass-throughs (`WarlockCheckboxRow`, `WarlockRadioButtonRow`) now call Jewel directly across 28 sites.

## Verification

`./gradlew check -PiosSkip=true -PlintSkip=true` passes; `:compose:compileKotlinIosSimulatorArm64` compiles.

**Not yet run on a device.** The gesture layering — tap-select vs long-press-drag vs strip scrolling, and whether the Initial-pass tracker sees the release before `onDragStopped` — is traced through the library sources but is timing-sensitive and wants a real device. `:androidApp:installDebug` fails with `INSTALL_FAILED_UPDATE_INCOMPATIBLE` against the emulator's differently-signed build, and uninstalling would wipe saved connections.

## Summary by CodeRabbit

- **New Features**
  - Added reorderable phone tabs with window selection, closing, settings, and activity indicators.
  - Added a window navigation rail for quickly adding and switching between game windows.
  - Added responsive phone and large-screen layouts.
  - Enabled edge-to-edge display with theme-aware system-bar icons.
  - Added improved keyboard handling and command-entry focus behavior.

- **Bug Fixes**
  - Improved window and tab state persistence across sessions.

- **Style**
  - Updated controls and dialogs for a more consistent visual experience.
  - Replaced the mobile navigation drawer with direct top-bar access to settings.


sproctor:
Thanks — five of the eight are fixed in f9ef2eb7. Two I'm declining, with reasons.

## Fixed

**Persisting a tab edit before the saved order loads** — real, and the worst of the batch since it silently destroys data. `editPhoneTabs` now returns while the order is `Loading`.

**Duplicate names in the decoded list** — real. Nothing we write can produce a repeat (`addPhoneTab` guards, `moveInSavedOrder` can't duplicate), but the value is a plain blob in the settings store, and the strip keys its `LazyRow` items by name, so a damaged one would crash rather than degrade. Deduplicated at the decode boundary, with a test.

**`width > Large.breakpoint`** — real off-by-one introduced when `WindowWidthSizeClass` was folded into `MobileGameLayout`. The size class it replaced treated 840dp as Expanded, so `>=` restores the old behaviour and matches the M3 spec.

**Scaffold padding discarded on the dashboard** — real. `DashboardView` takes a `modifier` and renders a full-size `Surface`, so its top content sat under `WarlockTopBar`.

**Settings icon has no accessible name** — real, `contentDescription = "Settings"`.

**Selection out of step with the strip's fallback** (nitpick) — worth doing. `currentTab` falls back to main when the selection leaves the strip, but the view model kept the stale name, so find-in-window and the selected styling pointed at a window no longer on screen. Now synced.

## Declining

**Remove `kotlin.native.enableKlibsCrossCompilation`** — the premise is wrong: this property is not newly added. It is present in the parent commit (`HEAD~1:gradle.properties` line 5); the diff only reordered the block, which made it look new. It is also plausibly load-bearing here, since this project's iOS klibs are cross-compiled from a Linux host. Removing a pre-existing build property on that basis is out of scope for this PR and carries a real risk of breaking the iOS build.

**Reorder `WindowView`'s parameters so `clearStream` precedes the defaulted ones** — the ordering is pre-existing and was preserved deliberately when the function became `expect`/`actual`; this PR did not introduce it. Changing it touches the `expect`, both `actual`s and every call site for no behavioural change, and every caller already passes it by name. Not worth the churn in a 63-file PR — reasonable as its own cleanup later.

`check -PiosSkip=true -PlintSkip=true` passes and the iOS simulator target compiles.

PR #302, opened 2026-08-21
Let a window with no saved settings inherit the base style

sproctor:
## The symptom

On a device with no settings yet, the story window renders light-on-dark whatever the theme. Everything around it is correctly light.

Measured from a light-theme tablet:

| element | rendered | skin light | skin dark |
|---|---|---|---|
| story body background | `#1E1F22` | `#FFFFFF` | `#1E1F22` |
| story body text | `#F0F0FF` | `#1E1F22` | `#F0F0FF` |
| `link` | `#2962FF` | **`#2962FF`** | `#82B1FF` |
| `roomName` background | `#1565C0` | **`#1565C0`** | `#82B1FF` |
| `bold` | `#D49200` | **`#D49200`** | `#FFFF00` |
| command entry background | `#FFFFFF` | **`#FFFFFF`** | `#1E1F22` |

Every preset resolves light, and so does the command entry — which reads the base style directly. Only the window body is dark, so this was never a dark-mode detection problem.

## The cause

`GameViewModel` built a window with no saved settings row as:

```kotlin
style = entity?.getStyle(colorPalette.value) ?: SAFE_DEFAULT_STYLE,
```

`WindowViewScaffold` then merges that over the base with `uiState.style.mergeWith(defaultStyle)`, and `mergeWith` keeps whatever `this` specifies. Both of `SAFE_DEFAULT_STYLE`'s colours are specified, so the safety net won outright and the resolved base style never reached the screen. `SAFE_DEFAULT_STYLE` is light-on-dark with no light variant, hence the result.

It only shows on a device with nothing saved — once a window has a settings row, `entity` is non-null and `getStyle` returns unspecified colours for anything the row does not set, which inherits correctly.

## The fix

- A window with nothing saved now starts as `StyleDefinition()` — all unspecified — so it inherits the base. Windows that do have a saved row are unaffected.
- `SAFE_DEFAULT_STYLE` moves to the **bottom** of the base cascade in `WindowRegistryImpl.baseStyle`, below `skin["default"]`. That is where it belonged: it exists for the case where no skin is loaded at all, and above the skin it could never lose. Legibility is still guaranteed if the skin fails to parse.

## Verification

`./gradlew check -PiosSkip=true -PlintSkip=true` passes; `:compose:compileKotlinIosSimulatorArm64` compiles.

Not yet confirmed on screen — `installDebug` fails against the emulator with `INSTALL_FAILED_UPDATE_INCOMPATIBLE` (differently-signed build), and uninstalling first would clear the device's saved connection. Expected result: a white story body with `#1E1F22` text, and `link`/`roomName`/`bold` unchanged from the table above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved window styling resolution so saved and inherited styles are preserved.
  * Added a fallback appearance when no skin or custom styling is available.
  * Ensured dynamically created windows receive consistent styling through the correct style hierarchy.
  * Updated the mobile game header to display the character’s game code consistently.
  * Improved color and style handling across skin, character, global, and legacy settings.

### DIFF
--- a/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/game/GameViewModel.kt
+++ b/compose/src/commonMain/kotlin/warlockfe/warlock3/compose/ui/game/GameViewModel.kt
@@ -68,7 +68,6 @@ import warlockfe.warlock3.compose.ui.window.WindowFindUiState
 import warlockfe.warlock3.compose.ui.window.WindowUiState
 import warlockfe.warlock3.compose.ui.window.getStyle
 import warlockfe.warlock3.compose.util.LatestValueWriter
-import warlockfe.warlock3.compose.util.SAFE_DEFAULT_STYLE
 import warlockfe.warlock3.compose.util.openUrl
 import warlockfe.warlock3.core.client.ClientCloseWindowEvent
 import warlockfe.warlock3.core.client.ClientCompassEvent
@@ -500,7 +499,9 @@ class GameViewModel(
             WindowUiState(
                 name = "main",
                 windowInfo = mutableStateOf(windows.value.firstOrNull { it.name == "main" }),
-                style = SAFE_DEFAULT_STYLE,
+                // Unspecified, not a concrete style: a window with nothing saved has to inherit the
+                // base style, and mergeWith keeps whatever this one specifies.
+                style = StyleDefinition(),
                 data =
                     StreamWindowData(
                         stream = windowRegistry.getOrCreateStream("main") as ComposeTextStream,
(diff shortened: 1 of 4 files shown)


sproctor:
Added in the latest commit, though not quite in the suggested form.

Testing `resolve()` on a hand-built list would only re-test `resolve`, and its first-set-wins behaviour is already covered by `StyleResolveTest`. More to the point, a test that builds its own list cannot notice the *production* list being reordered — which is exactly the regression here.

So the cascade is now a named function, `resolveBaseStyle`, lifted out of the `combine`, with `BaseStyleCascadeTest` over it:

- the skin's `default` beats the safety net
- the safety net applies only when no skin is loaded, or when the skin has no `default` preset
- character base beats global base beats skin
- a legacy `default` preset sits under its own scope's base, but still beats the scope below
- a skin-referenced colour (`textColorRef = "default"`) resolves against the skin palette — the mechanism that lets a character's base track the theme instead of freezing a literal

Confirmed the test has teeth by putting the safety net back above the skin: `theSkinDefaultBeatsTheSafetyNet` and `theCharacterBaseBeatsTheSkin` both fail, and pass again once the order is restored.

`check -PiosSkip=true -PlintSkip=true` passes and the iOS simulator target compiles.

PR #310, opened 2026-09-03
Tell the user when settings fail, and refuse to start on unreadable ones

sproctor:
Two Sentry issues, one dying disk on one machine a day apart. Both reached the thread's uncaught handler, which reported them as fatal to a service the user cannot read and told the user nothing.

- **DESKTOP-3X** — a *write*. `SQLITE_IOERR_FSYNC` out of a `COMMIT`, from a `viewModelScope.launch` nobody awaits. The app survived (Sentry's `fatal` is its own label; its handler delegates and never exits) and the setting was silently lost.
- **DESKTOP-3Y** — the *open*, once the volume had gone read-only. Room could not create the `.lck` file it locks while opening, and because Room builds lazily the `IllegalStateException` came out of the first query — `runBlocking { clientSettings.getWidth() }` — on the main thread, before a window ever appeared. The app did not start and said nothing about why.

## Writes report to the user

Every SQLite write already sat in a `withContext(NonCancellable)`. Those become `persistToDatabase`, which keeps the write open across cancellation and hands a failure to `SettingsProblems` instead of throwing. The app shell shows it, **once per episode** — a disk that has stopped taking writes fails every save there is, so without that, dismissing the report would only summon the next one; it stays quiet until a save succeeds again.

This is what the TOML side of preferences has always done. A settings write that goes to SQLite should not be able to take down what the identical write to a config file would only log.

## Unreadable settings stop the launch

`openPrefsDatabase` probes with a real query — which is the point, since that is what moves Room's lazy open somewhere it can be handled — and throws `SettingsUnreadableException`. Desktop shows a window naming the file and what the OS said, and exits.

Carrying on is not an option worth having: every store reads missing data as *"nothing saved yet"*, so an app that started anyway would come up with empty settings and then save them over settings that were only unreachable, not gone.

**The config stores had the same hole with worse consequences, and it is not in either report.** `TomlFileStore.readFile` and `CharacterConfigStore`'s per-section read both logged `"ignoring it"` and returned null, which every caller reads as "no file" — so a `highlights.toml` with a bad byte, or a typo in a file the docs call hand-editable, looked exactly like a fresh install, and the next save wrote that emptiness over it. Their startup reads now throw too. Only the startup reads: the write-path reads keep log-and-continue, since there is no launch left to abort and making them fatal would only be a new crash out of a click handler.

The reason shown is the **deepest cause, named by type**. Room's outermost message asks whether the developer configured a path correctly; three causes down sits `warlock-v22.db.lck: Read-only file system`, which is what actually tells the user what to fix. The type goes in front because a JVM `AccessDeniedException` for the same file carries only the path, and a bare path says nothing.

## Verified against the real app

Not just the tests — run with a read-only config directory:

```
Warlock cannot start
────────────────────
Warlock could not read your settings database.

AccessDeniedException: /home/sproctor/.config/warlock/warlock-v22.db.lck

Warlock will close rather than start with empty settings and save them
over what is on disk. Settings are kept in /home/sproctor/.config/warlock;
check that the files there are readable and undamaged.
                                                              [ Quit ]
```

**The first attempt at that window was broken and this caught it.** It reused `WarlockAlertDialog`, which is a `DialogWindow` — with no application window to own it, it drew nothing while `application` waited for a window to close, leaving the app hung and invisible: a quieter version of the bug being fixed. It is a plain top-level `Window` now, and closing it ends the process (confirmed with a real `WM_DELETE_WINDOW`).

An earlier run is worth noting too: with only the *directory* read-only but the `.lck` already present, the database opened fine and nothing fired. The probe does not produce false positives — which matters, since a false positive would wrongly refuse to start a working install.

## Tests

696 pass; `./gradlew check -PiosSkip=true -PlintSkip=true` green on a clean `--rerun-tasks`.

- `DatabaseWriteFailureTest` — each repository reports rather than throws; twenty failures produce one report; dismissal silences, a success re-arms; a write already in flight still completes when its caller is cancelled.
- `PrefsDatabaseOpenFailureTest` — an unwritable directory throws with the root cause, not Room's wrapper; a writable one round-trips across a reopen.
- `UnreadableConfigTest` — an unreadable *and* an unparseable `highlights.toml` both stop the load and leave the file untouched; a **missing** file is still just a fresh install, which is the distinction the whole change turns on.

## Not covered

On Android and iOS the exception propagates rather than being shown — that closes the app without a message, so no empty settings get saved over real ones, but the user gets the system crash dialog. A config read failure there is swallowed by the eager background init, since only the desktop entry point calls `initialize()` itself. Closing that gap means blocking Android's `onCreate` on config I/O, which seemed the wrong trade to make unasked.

https://claude.ai/code/session_01Fo5nR4mXCTMAdXL9kH5Gzn

## Summary by CodeRabbit

* **New Features**
  * Added clear alerts for settings that cannot be read or saved, including guidance for resolving startup problems.
  * Added persistence for character ignore lists and an option to ignore AppImage integration.
  * Improved AppImage desktop integration with installation, refresh, and failure reporting.

* **Bug Fixes**
  * Unreadable or inaccessible settings now prevent startup instead of appearing as empty.
  * Settings writes continue safely when the initiating operation is cancelled.
  * Repeated failures are consolidated, while later failures are reported after recovery.


sproctor:
Addressed all four findings — each verified against the code first, and two of them turned out to be real holes in what this branch set out to fix.

**`unreadable` could be erased before anyone read it** (AppContainer). Declared below the `init` block that starts `initialize()` on the IO dispatcher, and Kotlin runs initializers in textual order — so the constructor's `= null` raced the coroutine and could wipe the exception that stops the launch. Moved above the block, and now read under the same lock that writes it.

**`metadataOrNull` conflating "not there" with "cannot look"** — confirmed, and the real semantics are worse than the review supposed. I probed them against kotlinx-io rather than trusting the docs:

| case | `metadataOrNull(file)` | `list(parent)` | `source(file)` |
|---|---|---|---|
| file mode 000 | present | shows it | `FileNotFoundException (Permission denied)` |
| dir `r` no `x` | **null → "absent"** | shows it | Permission denied |
| dir mode 000 | null | **`[]` — no exception** | Permission denied |
| genuinely absent | null | absent | `FileNotFoundException (No such file…)` |
| missing parent | null | throws | throws |

So `list()` on an unenterable directory returns *empty* rather than throwing, and `FileNotFoundException` cannot discriminate. Presence now comes from the parent's listing, which answers directly and closes the stat-then-open race too. The one case nothing can answer — an unenterable directory being indistinguishable from an empty one — is documented where it's decided; such a directory denies writes as well, so the settings under it can't then be overwritten and the save failure is reported in its own right.

**The systemic half of that finding, which mattered more:** any failure reading the config files is now fatal, not only the `SettingsUnreadableException` the stores raise for a file they could identify as damaged. `initialize()` was swallowing everything else and coming up empty — the exact behaviour this branch exists to prevent, reachable by any path we hadn't thought of. Migration and macro seeding stay best-effort.

**The dismissal recovery boundary** — real. A report can sit on screen while writes recover and be dismissed minutes later; silencing then swallowed the next real failure. A success while a report is up marks it overtaken, and dismissing an overtaken report silences nothing.

**Tests assuming `setReadable(false)` obliged** — fair. They probe the denial and skip when it doesn't take (root, or a filesystem ignoring POSIX modes). CI is a normal ubuntu-latest runner so they still run there; nothing skipped locally.

Both new behaviours are pinned by tests I checked fail against the old code: failure → success → dismiss → failure, and a character directory with its search bit cleared. 698 tests pass, `check` green on a clean `--rerun-tasks`.

dd94561b

sproctor:
Valid — and it's the same conflation the helper was written to remove, so thanks for catching it. Fixed in c1de375a.

`entryNames` caught `IOException` wholesale and returned an empty set, i.e. a directory we cannot read reported as a directory with nothing in it. Only `FileNotFoundException` means "no such directory" now; every other listing failure is rethrown at startup and swallowed on the write path, where there is no launch left to abort and throwing would only be a crash out of a click handler. That's the same split the section reads already use, so the strictness is threaded through the callers rather than switched on globally — `readCharacterDir` is shared with `mutate`, while `isCharacterDir` is reached only from `load()` and is always strict.

One correction to the finding's premise, from measuring rather than reading the docs: on the JVM an inaccessible directory does **not** raise anything from `list()` — it returns an empty list. So that particular case isn't reachable through the catch either way, and the helper documents it as the one question no caller can answer (mitigated by the fact that a directory denying reads denies writes too, so the settings under it can't then be overwritten).

What *does* raise a non-`FileNotFound` `IOException` is a path that isn't a directory at all. So a settings folder that is somehow a regular file now stops the launch instead of silently starting with defaults and saving over it — that's the new test, and I verified it fails against the previous commit's wholesale catch.

699 tests pass, `check` green on a clean `--rerun-tasks`.

sproctor:
Fixed in da124126 — the structural reading is right, though the trigger isn't reachable yet and I couldn't write a test that fails without the fix. Details, since the second half matters:

**The valuable half is the flag, not the exception type.** `initialize()` sets `initialized = true` *before* the work so it runs exactly once. That's right for a failure it records and stops on, and wrong for cancellation: the flag stays set for an attempt that loaded nothing, so the entry point's own call returns at once with empty stores **and no error**, and the first save writes that emptiness over settings that were never gone. That's the exact outcome the rest of the branch exists to prevent. Cancellation is now rethrown with the flag cleared, and the best-effort migration below no longer logs cancellation as a config-store failure.

**No test, because none can fail.** Neither `load()` contains a suspension point, and Kotlin cancellation is cooperative — with nothing to suspend at, there's nothing for a cancel to throw at. I checked rather than assumed: a coroutine that cancels its own job and *then* calls both loads runs them to completion (`PROBE cancel-then-load -> completed normally`). I did write the racing cancel-then-retry test you asked for, and deleted it when it passed just as happily with the fix removed — a test that can't fail is worse than none, since it reads as coverage.

So this is a guard, not a repair. It becomes a live and silent bug the moment the load gets a suspension point — wrapping it in `withContext(ioDispatcher)` would do it, which is a natural enough change that the three lines seem worth keeping. I've left that change out of this PR as unrelated scope.

`check` green on a clean `--rerun-tasks`.

